PR #1, opened 2022-04-29
fix: Log debug level at service level

### LINKED ISSUE #33: Log and tag UDS traffic in UDSClient
Currently uds traffic is logged in the transports which causes code duplication. Move this to the UDSClient.

rumpelsepp:
Currently we rely on UDS traffic logged at the transport level. That's
an unfortunate layer violation and it is better to log this at
debug level at the service level instead.

Transports are degraded to log_trace(), since these data is only needed
when debugging is needed.


fkglr (commented):
Does this mean, that the information if something is a read or write is no longer present for the transport level in the trace? In that case debugging errors might become harder.

rumpelsepp:
Yes, you are right. We need to discuss this further. I talked to @peckto, it might be worth it to update the penlog spec.

PR #3, opened 2022-04-29
chore: Enable mypy strict setting

rumpelsepp:
mypy's strict mode is a collection of options which is maintained
upstream. This reduces the need to maintain the options on our side.

https://mypy.readthedocs.io/en/stable/config_file.html#confval-strict

### DIFF
--- a/src/cursed_hr/cursed_hr.py
+++ b/src/cursed_hr/cursed_hr.py
@@ -26,9 +26,9 @@ from pathlib import Path
 from typing import Any, BinaryIO, Optional, Union
 
 import zstandard as zstd
-from penlog import MessagePrio, RecordType
 
 from gallia.uds.core.service import NegativeResponse, UDSRequest, UDSResponse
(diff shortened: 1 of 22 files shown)


src/cursed_hr/cursed_hr.py
```diff
@@ -749,19 +749,19 @@ def next_sufficient_pointer(
         :param entry_id: The index of the entry, to which the given pointer points, in self.entries.
         :return: The next priority pointer's index along the corresponding priority zone.
         """
-        zone, pointer = self.next_sufficient_pointer_without_filters(
+        zone, ptr = self.next_sufficient_pointer_without_filters(
```
fkglr, line 752: Why has this variable been renamed?

rumpelsepp: mypy complains that the variable `pointer` (which is a function argument) is overwritten. IIRC the type differs, too.

rumpelsepp:
#248

PR #41, opened 2022-04-29
feat: Added dynamic parsing to scan-identifiers and scan-services

rumpelsepp:


peckto:
This seems already to be integrated into the `UDSClient.request` method:

https://github.com/Fraunhofer-AISEC/gallia/blob/master/src/gallia/uds/core/client.py#L79

https://github.com/Fraunhofer-AISEC/gallia/blob/master/src/gallia/uds/helpers.py#L101

rumpelsepp:
@fkglr can this be closed? See @peckto's comment.

fkglr:
> This seems already to be integrated into the `UDSClient.request` method:
> 
> https://github.com/Fraunhofer-AISEC/gallia/blob/master/src/gallia/uds/core/client.py#L79
> 
> https://github.com/Fraunhofer-AISEC/gallia/blob/master/src/gallia/uds/helpers.py#L101

Correct, the changes in this Pull Request have no effect anymore in the current code base.

PR #67, opened 2022-05-06
doc: Added more details to the setup.md

### LINKED ISSUE #14: Insufficient dependencies in docu
Original reporter: @ferdinandjarisch 

From what I see at least the following are missing [in the docu](https://fraunhofer-aisec.github.io/gallia/setup.html#dependencies):

* jq (penrun)
* dumpcap (über package tshark oder wireshark-common)

@fkglr @VeroSec @emedav Please add from your recent experiences…

Maybe one should also give instructions on how to install poetry, even if only a link to their [installation wiki](https://python-poetry.org/docs/#installation)

VeroSec:
Followed Issue https://github.com/Fraunhofer-AISEC/gallia/issues/14.

### DIFF
--- a/docs/setup.md
+++ b/docs/setup.md
@@ -25,17 +25,15 @@ $ cd gallia
 **penlog**
 
 See https://github.com/Fraunhofer-AISEC/penlog
+
 ```shell-session
-make
+$ make
 ```
 
 **penrun**
 
-Install `bin/penrun` into your PATH. 
-```shell-session
-export PATH="installpath/gallia/bin$PATH"
-```
-
+Install `bin/penrun` into `$PATH`.
+[`$HOME/.local/bin`](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) might be a good candidate.
 
 ### 3) Install gallia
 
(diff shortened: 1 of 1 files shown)


docs/setup.md
```diff
@@ -27,13 +27,22 @@ $ cd gallia
 **penlog**
 
 See https://github.com/Fraunhofer-AISEC/penlog
+```shell-session
+make
```
rumpelsepp:
```
$ make
```

rumpelsepp: Prefix with a `$` in order to indicate that this is an interactive shell session.

docs/setup.md
```diff
@@ -27,13 +27,22 @@ $ cd gallia
...
+```
 
 **penrun**
 
 Install `bin/penrun` into your PATH. 
+```shell-session
+export PATH="installpath/gallia/bin$PATH"
```
rumpelsepp: `$` prefix; see above.

rumpelsepp (changes requested):

rumpelsepp:
We should also mention in the `contributing` documentation that we follow the [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) specification. Please just prefix your commit message with `doc:`; I created #68. :)

docs/setup.md
rumpelsepp: The `export` line is wrong. The components of the `PATH` variable need to be separated by colons `:`. It is IMO a bad idea to instruct users to modify their `PATH`. We could either rephrase this and link to e.g. the [POSIX standard](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap08.html#tag_08_03) where the `PATH` variable is documented. 

Or the lazy path: Just leave it as as, just mention that `penrun` should live somewhere in `PATH`. -> `penrun` will be replaced by a python implementation in [v1.1.0](https://github.com/Fraunhofer-AISEC/gallia/milestone/2) rendering this section obsolete.

rumpelsepp (approved):

PR #77, opened 2022-05-09
Rework transport arguments with generic types

rumpelsepp:


rumpelsepp:
This one here brings the solution for all our discovery problems. Stay tuned … I will break everything. :D

rumpelsepp:
#249

Issue #81, opened 2022-05-09
(Sub)Command structure


rumpelsepp:
I hit the “Create Issue from Discussion Button”, since discussion could not be linked into projects.

### Discussed in https://github.com/Fraunhofer-AISEC/gallia/discussions/76

<div type='discussions-op-text'>

<sup>Originally posted by **rumpelsepp** May  9, 2022</sup>
I worked a bit on #9 and reached a blocker. Gallia is going to become a framework which is not limited to uds only. Therefore, it make sense to structure the CLI interface better. The challenge is to create a somehow self describing CLI interface which is easy to remember and easy to use.

I had a few ideas for structuring the commands. The structure of the code base could follow this accordingly. Personally, I think “Subcommands grouped by operation” is cool.

## Subcommands grouped by service

* `gallia`
  * `uds`
    * `scan`
      * `identifiers`
    * `discover`
    * `primitive`
      * `dtc`
      * `read-data-by-identifier`
  * `xcp`
    * `discover`
  * `dlt`
    * `discover`

### Examples

```
$ gallia uds scan identifiers
$ gallia xcp discover
$ gallia uds primitives dtc
```

## Subcommands grouped by operation

* `gallia`
  * `scan`
    * `uds`
      * `identifiers`
  * `discover`
    * `xcp`
    * `uds`
    * `dlt`
  * `primitives`
    * `uds`
      * `dtc`
      * `read-data-by-identifier`

### Examples

```
$ gallia scan uds identifiers
$ gallia discover xcp
$ gallia primitives uds dtc
```

## No subcommands grouped by prefix

* `gallia`

  * `dlt-discover`
  * `uds-discover`
  * `uds-primitives-dtc`
  * `uds-primitives-read-data-by-identifier`
  * `uds-scan-identifiers`
  * `xcp-discover`

### Examples

```
$ gallia uds-scan-identifiers
$ gallia xcp-discover
$ gallia uds-primitives-dtc
```</div>

PR #84, opened 2022-05-10
feat: Add dnsmasq wrapper script

rumpelsepp:
This was adapted from my bash script here:
https://codeberg.org/rumpelsepp/netscripts/src/branch/master/dhcp-and-router.sh

Open tasks:

* [ ] Create `GalliaBaseSync` and `GalliaBaseAsync`, #86 
* [ ] Create Subcommands, #81 
* [ ] Make a more basic `GalliaBase` without dumpcap, database logging and everything. #85 
* [ ] Prevent spamming the journal; there is a dnsmasq flag

### DIFF
--- a/src/gallia/commands/dhcp.py
+++ b/src/gallia/commands/dhcp.py
@@ -1,0 +1,168 @@
+from argparse import Namespace
+from subprocess import run
+
+from gallia.udscan.core import GalliaBase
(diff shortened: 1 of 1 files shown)


rumpelsepp:
how to dnsmasq should only live in the docs

PR #95, opened 2022-05-12
fix(db): correct index declaration

fkglr:
Currently the initialization of a new database fails, because the index declaration is wrong.


rumpelsepp (approved):
Sorry, my bad… :see_no_evil:

PR #116, opened 2022-05-20
feat: Add a Script class

### LINKED ISSUE #85: Create a more basic `GalliaBase`
`GalliaBase` includes a lot of stuff. For just adding commands to `gallia` it is a bit too much. Create a very basic `GalliaBase` class.

### LINKED ISSUE #86: Separate `GalliaBaseSync` from `GalliaBaseAsync`
#85 suggest providing a more basic `GalliaBase` class. In several use cases it is not needed to provide an `async def main(…) …`. Provide a base class which is not async.

rumpelsepp:
This adds a script class to gallia. With a Script/AsyncScript it is
possible to add commands to gallia which do not need advanced features,
such as database logging or artifacts dirs.

Code which is related to scanners was moved to the Scanner class.

### DIFF
--- a/src/gallia/udscan/core.py
+++ b/src/gallia/udscan/core.py
@@ -53,7 +53,16 @@ class Formatter(ArgumentDefaultsHelpFormatter):
 
 
 class GalliaBase(ABC):
-    """Base class for all gallia tasks"""
+    """GalliaBase is a baseclass for all gallia commands.
+    In order to register cli arguments:
+
+    - `add_class_parser()` can be overwritten to create
+       e.g. scanner related arguments shared by all scanners
+    - `add_parser()` can be overwritten to create specific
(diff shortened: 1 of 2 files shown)


src/gallia/udscan/core.py
```diff
@@ -53,36 +53,197 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
...
+
+        try:
+            self.main(args)
+            return 0
+        except KeyboardInterrupt:
+            return 128 + signal.SIGINT
+
+
+class AsyncScript(GalliaBase, ABC):
+    """Script is a base class for a asyncronous gallia command.
```
peckto:
```suggestion
    """AsyncScript is a base class for a asyncronous gallia command.
```

peckto (approved):

PR #136, opened 2022-05-27
feat(cli): Unified output

fkglr:
Currently the output on the cli / in the logs is a mess.
E.g. for scan-sessions, sessions are sometimes printed in hex, sometimes in decimal format, sometimes in hex but as if they were decimal, which leads to max confusion:

image

This PR aims to unify the output and does so, by offering a simple method `g_repr()`, which can be called for any object to be formatted in universal style. Should not be called for cases where a specific formatting (e.g. always decimal) is intended.
This method also handles lists and dicts, so there is no need to either fall back to default python formatting or to write extra code for such cases anymore.
With these changes applied, the same scan run looks as follows:

image

### DIFF
--- a/src/gallia/db/db_handler.py
+++ b/src/gallia/db/db_handler.py
@@ -10,6 +10,7 @@ import aiosqlite
 from gallia.uds.core import service
 from gallia.uds.core.utils import bytes_repr as bytes_repr_
 from gallia.penlog import Logger
+from gallia.utils import g_repr
 
 
 def bytes_repr(data: bytes) -> str:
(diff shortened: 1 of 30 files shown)


rumpelsepp (approved):

rumpelsepp:
Pls merge yourself once it is finished! ACK

fkglr:
It might be, that not each and every output string is already covered. Especially in the lib files, there are a few cases that I excluded to avoid cyclic redundancies.

fkglr:
I also removed some of the unnecessary str() statements for parameters passed to a logging function, because the logging function already calls str() on said parameters internally.

PR #154, opened 2022-06-10
Payload fuzzer

VeroSec:
Added a scanner in which the payload of the service WriteDataByIdentifier can be fuzzed. 
It is possible to scan multiple services and data identifiers for the same ECU.
Additionally, an asynchronous task to observe specific CAN massages can serve as a feedback channel while fuzzing to look at possible message changes, that occurred because of the fuzzing process.

### DIFF
--- a/src/gallia/udscan/scanner/fuzz_payloads.py
+++ b/src/gallia/udscan/scanner/fuzz_payloads.py
@@ -1,0 +1,204 @@
+# SPDX-FileCopyrightText: AISEC Pentesting Team
+#
+# SPDX-License-Identifier: Apache-2.0
+
+import asyncio
+import binascii
+import random
+from argparse import Namespace
+
+from gallia.transports.base import TargetURI
+from gallia.transports.can import RawCANTransport
+from gallia.uds.core.client import UDSErrorCodes, UDSRequestConfig
+from gallia.uds.core.exception import IllegalResponse
+from gallia.uds.core.service import NegativeResponse, UDSResponse
+from gallia.uds.helpers import suggests_identifier_not_supported
+from gallia.udscan.core import UDSScanner
+from gallia.utils import auto_int
+from gallia.uds.core.constants import UDSIsoServices
+
+
+class FuzzPayloads(UDSScanner):
+    def __init__(self) -> None:
+        super().__init__()
+        self.can_recv_event = asyncio.Event()
+
+    def add_parser(self) -> None:
+        self.parser.add_argument(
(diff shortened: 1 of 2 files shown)


rumpelsepp:
I took the liberty and fixed all the ci-checks. The ci was mainly confused because of two problems which I fixed in master (8c0dd728a38fe766c5fc7e576becb53d1cf556f1, 382bd5e5baddeb867b90032eb359e543421dc1af).

Be careful to use unix fileendings instead of windows fileendings. There is a [git setting](https://git-scm.com/docs/git-config#Documentation/git-config.txt-coreautocrlf) that helps dealing with this problems automatically; I have this in my `~/.config/git/config` file:

```
[core]
	autocrlf = input
```

src/gallia/udscan/scanner/fuzz_payloads.py
```diff
@@ -0,0 +1,192 @@
...
+        return random.randbytes(random.randint(args.min_length, args.max_length))
+
+    async def observe_can_messages(self, can_ids: list[int], args: Namespace) -> None:
+
+        can_url = args.target.url._replace(scheme=RawCANTransport.SCHEME)
+        transport = RawCANTransport(TargetURI(can_url.geturl()))
+        await transport.connect()
+        transport.set_filter(can_ids, inv_filter=False)
+
+        self.can_recv_event.set()
```
rumpelsepp: The event is never read in the main loop, isn't it?

rumpelsepp:
Took the liberty to rebase against master und fix the linters. I also added @emedav as a Co-Author. Please pull with:

```
$ git pull --rebase
```

Is this ready to merge?

src/gallia/udscan/scanner/fuzz_payloads.py
VeroSec: The Event is not read, but it triggers a log message if the can reply changed.

VeroSec (commented):
reviewed and resolved all conversations

src/gallia/udscan/scanner/fuzz_payloads.py
rumpelsepp: No, it doesn't. The event is not read, thus it triggers nothing. The log message is generated in a background task which is created in line 123. I suggest removing the event.

rumpelsepp (changes requested):
Please remove the unused event. Otherwise looks good to me. I added the required Copyright and License tags. Now the CI runs.

PR #160, opened 2022-06-14
ci: Add more simple scripts to CI tests

peckto:


peckto:
We first need to configure the vECU to support read and clear DTC.

rumpelsepp (approved):

PR #162, opened 2022-06-14
Basic DTC response functionality for the vECU

fkglr:
Adds basic functionality for ClearDiagnosticInformation and ReadDTCInformation (only sub function reportDTCByStatusMask). The former has no effect.


peckto (approved):

fkglr:
During testing I noticed, that there is a bug in the handling of subfunction availability. More precisely, serviceNotSupportedInActive session is returned in cases where it is listed as available in the startup overview.
I would suggest to hold this request back until i have finally investigated if this is a consequence of this MR (unlikely) or is due to already existing code.

PR #172, opened 2022-06-20
fix: Behave as defined in the iso-standard and start in session 0x01

### LINKED ISSUE #140: strange race condition in scanners
Sometimes the vECU tests fail:

* https://github.com/Fraunhofer-AISEC/gallia/runs/6649856371?check_suite_focus=true
* https://github.com/Fraunhofer-AISEC/gallia/runs/6649776374?check_suite_focus=true
* https://github.com/Fraunhofer-AISEC/gallia/runs/6659074724?check_suite_focus=true

Find out why this happens.

rumpelsepp:
An attempt to fix #140

### DIFF
--- a/src/gallia/uds/ecu.py
+++ b/src/gallia/uds/ecu.py
@@ -101,6 +101,8 @@ class ECU(UDSClient):
         Returns:
             True on success, False on error.
         """
+        await self.diagnostic_session_control(0x01, config=config)
+        return True
 
     async def set_session_post(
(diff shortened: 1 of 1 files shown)


fkglr:
This might break some current code and lead to confusing bugs when chaining session transitions with set_session(). Furthermore, it can lead to unexpected differences between ECUs which override the set_session_pre method and those which do not.
Therefore, I think this should not be merged too quickly without properly discussing side effects and potential alternatives.

PR #174, opened 2022-06-20
fix: Let wait_for_ecu() recognize the cli --timeout value

### LINKED ISSUE #166: wait for ECU timeout should use cli value

rumpelsepp:


peckto:
We have a hard coded sleep in the `_wait_for_ecu` function:
https://github.com/Fraunhofer-AISEC/gallia/blob/a86d67b71ba89ffbfcdd23bee424c586b14ea5ab/src/gallia/uds/ecu.py#L325-L330

This will cause problems with timeout values < 1sec

rumpelsepp:
@peckto fixed.

peckto (approved):

PR #180, opened 2022-06-24
feat(scan-sessions): add fast search option

fkglr:
Adds the option `--fast` which still finds any (session A, session B) transition that the current version finds, but leaves out many of the other (mostly useless) stacks.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -73,6 +73,7 @@ jobs:
     - name: Test scan-sessions
       run: |
         poetry run gallia scan-sessions --target "tcp-lines://127.0.0.1:20162" --no-dumpcap --depth 2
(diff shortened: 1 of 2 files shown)


peckto:
Please add a test-case for the new option:
https://github.com/Fraunhofer-AISEC/gallia/blob/a86d67b71ba89ffbfcdd23bee424c586b14ea5ab/.github/workflows/tests.yml#L73-L75

fkglr:
> Please add a test-case for the new option:
> 
> https://github.com/Fraunhofer-AISEC/gallia/blob/a86d67b71ba89ffbfcdd23bee424c586b14ea5ab/.github/workflows/tests.yml#L73-L75

I've added a small test case similar to the main one for scan-session. A depth limitation is in my opinion not necessary here, as the scanner runs fast enough.

rumpelsepp (approved):

rumpelsepp:
scheduled for v1.0.2 as well, since another patch depends on this.

Issue #17, opened 2022-04-29
Log session transitions in DB


rumpelsepp:

mich41v4294:
Ready to be closed? Fixed by #182?

PR #182, opened 2022-06-24
feat(db): store sessions in the database

fkglr:
Adresses #17 

The idea behind this pull request is to remove the currently in place session scans at the start of most of the scanners and to replace them with the results from a full session scan as delivered by the scan-sessions script.
Therefore, the corresponding information is stored in a separate database table.

There is however still work to do here.
Currently only the logging to the database and the usage of this information during a session change is implemented and it is not clear if the current format, which can only stores transitions but no other information, is sufficient. Or in other words, can a broader format add enough value to justify replacing the simple session transition format?


rumpelsepp:
Increment Database Scheme Number? -> `1.0` -> `2` (let's drop the float)

PR #185, opened 2022-06-27
Fix CAN Filter to deal with exended ids

rumpelsepp:


src/gallia/udscan/scanner/find_can_ids.py
```diff
@@ -164,6 +164,8 @@ async def main(self, args: Namespace) -> None:
 
         self.logger.log_summary(f"Found {len(idle_msgs)} CAN Messages on idle Bus")
         transport.set_filter(idle_msgs, inv_filter=True)
+        # flush receive queue
+        await transport.get_idle_traffic(2)
```
rumpelsepp: I think we should fix this better. Why is this flushing needed?

src/gallia/transports/can.py
```diff
@@ -293,126 +295,104 @@ class RawCANTransport(BaseTransport, scheme="can-raw", spec=spec_can_raw):
...
         assert target.hostname is not None, "empty interface"
         self.interface = target.hostname
         self._sock: s.socket
-        self.src_addr: int
-        self.dst_addr: int
 
     async def connect(self, timeout: Optional[float] = None) -> None:
         self._sock = s.socket(s.PF_CAN, s.SOCK_RAW, s.CAN_RAW)
+        self._sock.setsockopt(s.SOL_CAN_RAW, s.CAN_RAW_LOOPBACK, 0)
+        self._sock.setsockopt(s.SOL_SOCKET, s.SO_TIMESTAMPNS, 1)
```
peckto: Unfortunately, python seems not to offer the `SO_TIMESTAMPNS` constant.

rumpelsepp:
@peckto Look again; I scheduled this into the next patch release as well.

peckto:
Should we filter all extended IDs when not scanning in extended-id mode and the other way around?
Mixing up endpoints with normal IDs and extended IDs does not make sense to me anyway and the generated target URL will be flawed.

rumpelsepp:
@peckto Ready

ferdinandjarisch:
Was implemented by another MR in the meantime.

PR #193, opened 2022-06-30
fix(uds): fix error code for failurePreventsExecutionOfRequestedAction

fkglr:


rumpelsepp:
This might help for the future: https://docs.python.org/3/library/enum.html#enum.unique

rumpelsepp (approved):

PR #194, opened 2022-06-30
feat: Add helper for lazy imports

### LINKED ISSUE #11: Implement Lazy Loading
``` python
import importlib.util
import sys


def lazy_import(name):
    if name in sys.modules:
        return sys.modules[name]
    spec = importlib.util.find_spec(name)
    loader = importlib.util.LazyLoader(spec.loader)
    spec.loader = loader
    module = importlib.util.module_from_spec(spec)
    sys.modules[name] = module
    loader.exec_module(module)
    return module
```

rumpelsepp:


rumpelsepp:
* https://peps.python.org/pep-0690
* https://discuss.python.org/t/pep-690-lazy-imports/15474

PR #195, opened 2022-06-30
fix: Raise a TimeoutError when wait_for_ecu() times out

rumpelsepp:
Commit 8f225c7422599b2ce3a10b4c2d529608028ba6f7 removed the return value of `wait_for_ecu()`. This brakes a gallia plugin which we use internally. I think it makes more sense to just raise the `TimeoutError`. @fkglr any thoughts?


rumpelsepp:
I thought about this. Rescheduled this for v1.1.0 for discussion. I will revert to the old behaviour in v1.0.3.

rumpelsepp:
`wait_for_ecu()` is currently being refactored by @ferdinandjarisch in his activities around #751. Closing this here.

PR #196, opened 2022-06-30
fix DoIP crashes and invalid URLs

rumpelsepp:
- fix(doip): Fix crash when DoIP gateway denies routing
- fix(doip): Generate the DoIP URLs in the correct order


rumpelsepp:
I have tested this live. Works and fixes a crash.

PR #205, opened 2022-07-05
feat: analyzer

peckto:
Adding a draft version of the gallia analyzer to analyze and visualize scan results.
Credit goes to @chanyoungso 

Example usage:
```
$ poetry install --extras analyzer
$ poetry shell
$ gallia analyze -aer --source gallia.db
```

Open points:
- [x] Load analyzer as optional gallia module to reduce loading time
- [x] use penlog logger
- [x] #221 
- [ ] docs
- [ ] create mapping between `vendor_lookup` table and ecu


pyproject.toml
```diff
@@ -32,6 +32,10 @@ zstandard = ">=0.17,<0.19"
 python-can = "^4.0.0"
 tabulate = "^0.8.9"
 construct = "^2.10.68"
+numpy = "^1.21.4"
+openpyxl = "^3.0.9"
+pandas = "^1.3.4"
+matplotlib = "^3.4.3"
```
rumpelsepp, line 57: I would suggest that we add this as extra dependencies and make the analyzer optional. Otherwise these heavy dependencies are always present.

rumpelsepp: https://python-poetry.org/docs/pyproject/#extras

peckto: Good idea.
For the cli interface we need to know, if we can add the Analyzer to the cli or not.
Is there a convenient way to detect if the packet was installed with `extra`?
Or do we need something like
```py
try:
    from gallia.analyzer import Analyzer
    load_module(Analyzer)
except ModuleNotFoundError:
    print("Install extra dependencies to use all gallia features")
```

peckto: done

peckto:
Because the information in the tables `vendor_lookup` and `vendor_session` are ECU specific and we can have results from multiple ECUs in one database, we need a mapping between the vendor specific information and the ECU.

To implement this mapping, we have two possibilities:
1) Create a "vendor_lookup_id" column in the ECU table (1:Vendor Lookup -> n:ECUs)
2) Create a "cross_vendor_lookup_ecu" table (n:Vendor Lookup -> m:ECUs)

My first guess would be (2), this is more flexible and we do not need to store vendor/analyzer specific information in the ECU table.

rumpelsepp:
Poetry 1.2 is out. Maybe this helps with making the analyzer optional:

https://python-poetry.org/docs/pyproject/#scripts

peckto:
rebased to current master

rumpelsepp:
Pls. reopen this if it gets relevant again.

PR #209, opened 2022-07-06
add support for TOML config

rumpelsepp:
Depends on #9

An optional config file can be created in the following locations; the first one which is found will be used. It is searched in these locations in this particular order.

* `$PWD/gallia.toml`
* `$current_git_root/gallia.toml`
* `~/.config/gallia/gallia.toml` (`$XDG_CONFIG_HOME` is recognized according to the [XDG standard](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html))
* `/etc/xdg/gallia/gallia.toml`

A path can be enforced by setting the environment variable `GALLIA_CONFIG` (the env variable avoids a chicken and egg problem with the `ArgumentParser`).

The config is used **only** for setting the **defaults** for the cli args. The cli flags have the highest priority, all config values have a corresponding flag. The `argparse.Namespace` object will stay as the source of truth for all configuration values. I suggest namespacing the configuration values.

A config template can be created like this:

```
$ gallia --template
```

### DIFF
--- a/src/gallia/command.py
+++ b/src/gallia/command.py
@@ -5,7 +5,6 @@
 import argparse
 import asyncio
 import json
-import os
 import signal
 import sys
 import time
@@ -20,7 +19,7 @@ from importlib.metadata import EntryPoint, entry_points, version
 from pathlib import Path
 from secrets import token_urlsafe
 from tempfile import gettempdir
-from typing import Optional, cast
+from typing import Any, Optional, cast
 
 import aiofiles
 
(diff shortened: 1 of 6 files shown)


src/gallia/command.py
```diff
@@ -212,13 +213,13 @@ def add_class_parser(self) -> None:
         group = self.parser.add_argument_group("generic gallia arguments")
         group.add_argument(
             "--artifacts-dir",
-            default=os.environ.get("PENRUN_ARTIFACTS"),
+            default=self.config.get("gallia.scan.artifacts_dir"),
```
peckto: Should we use `this.CATEGORY` instead of hard-coding `scan`?

rumpelsepp: done.

rumpelsepp:
@peckto It is done.

here you can see it in action: 735ff48f27e222f9db485ca1e466be9c573abb4c

Show all defaults for all commands:
```
$ gallia --show-defaults
```

Print a template with all config settings, commented.
```
$ gallia --template
```

If you create a config file and set a value there, you should see the default value changing.

Example:

```
$ cat gallia.toml
[gallia]
[gallia.scan]
target = "foo"
$ gallia --show-defaults | grep "foo"
```

PR #218, opened 2022-07-11
fix(ecu state): better mapping of responses to state changes

fkglr:
better handling of state change tracking in the ECU and adopted and simplified state changing in the virtual ECU.

Main change:
- session change resets the state, which better fits the UDS standard
- security access levels are only set on a successful session change not anymore on successful seed requests


rumpelsepp (approved):

rumpelsepp:
@fkglr Does this need to be ported to `1-0-maint`and thus the next patch release? If that's the case, please add the `maint` label and I will take care of this.

rumpelsepp:
maint: e674daea2fb55d9f0c0f64573b75df6f432d38fa

PR #220, opened 2022-07-15
fix(db): Update ecu state after reconnect

peckto:
After reconnect, the ECU might have lost its state (eg. session).
This PR updates the state after the reconnect.

### DIFF
--- a/src/gallia/uds/core/client.py
+++ b/src/gallia/uds/core/client.py
@@ -42,7 +42,8 @@ class UDSClient:
         self.logger = Logger("uds", flush=True)
 
     async def reconnect(self, timeout: Optional[int] = None) -> None:
-        """Calls the underlying transport to trigger a reconnect"""
+        """Calls the underlying transport to trigger a reconnect.
+        Be aware, that this might change the state of the ECU! Always restore desired state after reconnect."""
         await self.transport.reconnect(timeout)
 
     async def _read(
--- a/src/gallia/udscan/scanner/fuzz_payloads.py
+++ b/src/gallia/udscan/scanner/fuzz_payloads.py
@@ -5,6 +5,7 @@
 import asyncio
 import binascii
 import random
+import sys
 from argparse import Namespace
 
 from gallia.transports.base import TargetURI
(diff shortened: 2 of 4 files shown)


src/gallia/uds/ecu.py
```diff
@@ -382,7 +386,10 @@ async def refresh_state(self, reset_state: bool = False) -> None:
         if reset_state:
             self.state.reset()
 
-        await self.read_session()
+        try:
+            await self.read_session()
+        except UDSException as e:
+            self.logger.log_warning(f'Could not read current session as part of refresh_state: {e}')
```
peckto: Should we reset the state in this case?

peckto:
The error handling is not so trivial for this case.
We should keep in mind, that we have not only the state logged into the db (which is stored in the ecu), but also the internal state of the scanner, which is implicit.
If we change the state in the ecu, without notifying the scanner, we are in trouble.
For this reason, I think we should not change the state from inside the `reconnect` function.
Only the scanner itself should change the state.
This means, that if the state has changed because of the reconnect, we need to propagate the error back to the scanner.
The scanner should thus expect the reconnect to fail, not only because of a connection loss, but also due to a change in the ECU state.

Unfortunately, this can cause quite deep try-except blocks:
```py
try:
    ecu.do_something()
except Error:
    try:
        ecu.reconnect()
    except ConnectionError:
        ecu.power_cycle()
        try:
            ecu.wait_for_ecu()
        except Error:
            # Fatal Error
            sys.exit(1)
    except StateChanged:
        try:
            ecu.set_session(session)
        except Error:
            # Fatal Error
            sys.exit(1)
```

A related problem is, that we currently don't have higher level exceptions.
When a request fails, we typically reraise the last UDS Exception.
For the `reconnect` this means, that in case the `read_session` fails, it would raise a UDS `RequestOutOfRange` exception.
This is not intuitive and probably not what the user expects.
One solution would be to convert the `reconnect` function to type `bool` and communicate the error via the return code.
We could do this for the "changed state" case, but what about the "connection error" case?
I'm not sure, if it's a good idea, to mix return code and exceptions for error handling.

This is all not so convincing I must admit.
Does anyone have an idea how to solve this nicely?

peckto:
The problematic error handling maybe shows, that it's not a good idea to combine the update state with the reconnect function.
Maybe it's easier if the scanner is responsible to restore the desired state.

I updated this PR and added `check_and_set_session` after reconnect where its necessary.

rumpelsepp:
@peckto is this still relevant?

rumpelsepp:
Closing this. @fkglr if it is still relevant, please reopen.

PR #235, opened 2022-08-25
meta file

rumpelsepp:
- fix: Raise an exception when the connection is lost
- feat: Silence the logs of sqlite
- fix: Fix accidently renamed variables
- feat: Add support for META file

### DIFF
--- a/src/gallia/command.py
+++ b/src/gallia/command.py
@@ -356,7 +356,10 @@ class Scanner(BaseCommand, ABC):
 
     CATEGORY = "scan"
     ARTIFACTSDIR = True
-    CATCHED_EXCEPTIONS: list[type[Exception]] = [BrokenPipeError, ConnectionRefusedError]
+    CATCHED_EXCEPTIONS: list[type[Exception]] = [
+        BrokenPipeError,
+        ConnectionRefusedError,
+    ]
 
     def __init__(self, parser: ArgumentParser, config: dict[str, Any]) -> None:
         super().__init__(parser, config)
--- a/src/gallia/command.py
+++ b/src/gallia/command.py
@@ -122,7 +122,7 @@ class BaseCommand(ABC):
     EPILOG: Optional[str] = None
 
     LOGGER_NAME = "gallia"
-    ARTIFACTSDIR: bool = False
+    HAS_ARTIFACTS_DIR: bool = False
     CATCHED_EXCEPTIONS: list[type[Exception]] = []
 
     def __init__(self, parser: ArgumentParser, config: dict[str, Any]) -> None:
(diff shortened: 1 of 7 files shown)


src/gallia/command.py
```diff
@@ -226,9 +256,45 @@ def prepare_artifactsdir(
...
+            exit_code = 128 + signal.SIGINT
+        # Ensure thet META.json gets written in the case a
+        # command calls sys.exit().
+        except SystemExit as e:
+            exit_code = e.code
+        except Exception:
+            exit_code = ExitCodes.UNHANDLED_EXCEPTION
+            traceback.print_exc()
+        else:
+            self.logger.info(f"Stored artifacts at {self.artifacts_dir}")
```
peckto: Maybe we can move this to `finally`?

src/gallia/command.py
```diff
@@ -226,9 +256,45 @@ def prepare_artifactsdir(
...
+        finally:
+            if self.ARTIFACTSDIR:
+                self.run_meta.exit_code = exit_code
+                self.run_meta.end_time = datetime.now(tz).isoformat()
+                data = msgspec.json.encode(self.run_meta)
```
peckto, line 299: Does `msgspec` support pretty printing?

src/gallia/command.py
```diff
@@ -226,9 +256,45 @@ def prepare_artifactsdir(
...
+
+        exit_code = 0
+        try:
+            exit_code = self.run(args)
+        except (BrokenPipeError, ConnectionRefusedError) as e:
+            exit_code = 1
+            self.logger.critical(g_repr(e))
+        except KeyboardInterrupt:
+            exit_code = 128 + signal.SIGINT
+        # Ensure thet META.json gets written in the case a
```
peckto:
```suggestion
        # Ensure that META.json gets written in the case a
```

src/gallia/command.py
rumpelsepp: I do not like it, since in every case this line is printed. That's pretty annyoing tbh.

src/gallia/command.py
```diff
@@ -226,9 +256,45 @@ def prepare_artifactsdir(
...
+                self.get_file_log_level(args),
+                self.artifacts_dir.joinpath("log.json.zst"),
+            )
+        else:
+            setup_logging(self.get_log_level(args))
+
+        exit_code = 0
+        try:
+            exit_code = self.run(args)
+        except (BrokenPipeError, ConnectionRefusedError) as e:
```
rumpelsepp: @peckto Should we make this list configurable and provide these two as a default within gallia?

src/gallia/command.py
rumpelsepp: Yeah, looks a bit like what `@dataclass` does.

```
>>> class Dog(Struct):
...    name: str
...    breed: str
...    is_good_boy: bool = True
Dog('snickers', breed='corgi')
```

PR #256, opened 2022-09-15
feat: More flexibility for loading config files

ghost:
Allows to specify an additional list of paths in which to search for the configuration file. Might be especially useful for gallia scripts.


rumpelsepp:
Any idea for a shorter name? `additional_searchpath` is rather long. Apart from that: :+1:

Issue #90, opened 2022-05-10
Scan dump seeds should inform about progress after time rather than after modulo seeds


rumpelsepp:
When dumping seeds is real slow, one has to wait for an eternity to get the feedback of actual dumping speed.

/cc @ferdinandjarisch

PR #261, opened 2022-09-21
fix: change log output of seeds-dump to every 30s rather than 1kB

ferdinandjarisch:
Implements #90


rumpelsepp:
Yeah, we can do this. Master solution would be some progress indicator which is decoupled from the logging.

rumpelsepp (approved):

rumpelsepp:
Just merge it once the linters are green. :dancers:

PR #277, opened 2022-10-19
feat: Add lockfile support as known from penrun

rumpelsepp:


src/gallia/command.py
```diff
@@ -287,8 +288,20 @@ def prepare_artifactsdir(
 
     def entry_point(self, args: Namespace) -> int:
         if (p := args.lock_file) is not None:
-            self._lock_file_fd = os.open(p, os.O_CREAT)
-            fcntl.flock(self._lock_file_fd, fcntl.LOCK_EX | fcntl.LOCK_NB)
+            p.parent.mkdir(parents=True, exist_ok=True)
```
rumpelsepp: This should be put into an own method. `entry_point()` is already too large.

rumpelsepp: I mean the whole block, not only the line. Sorry. :D

PR #281, opened 2022-10-20
fix(cursed-hr): Handle penlog priority prefix

fkglr:


rumpelsepp:
As a temporary workaround this is entirely fine. The longterm fix is migrating to [`gallia.log.PenlogReader`](https://github.com/Fraunhofer-AISEC/gallia/blob/4f4b3793e9627c42c5a8d907ac9684e8cbf924df/src/gallia/log.py#L422).

PR #284, opened 2022-10-20
chore: Add stale bot

rumpelsepp:
Only issues without milestone or without assignee are handled. Issues
and PRs are notified after 5 months and closed after 6 months. Seems
fair for me.


rumpelsepp:
@Fraunhofer-AISEC/pin

PR #296, opened 2022-10-27
feat: Add progressbar support

rumpelsepp:
`gallia scan uds identifiers` is the first scanner with a progressbar.


rumpelsepp:
superseeded by #643

PR #310, opened 2022-10-31
Revert "fix: Only wait in the error branch"

rumpelsepp:
This reverts commit 8fb3ff1b6769720a88a934464536c5506609a033.


rumpelsepp:
This fix breaks `wait_for_ecu()` for my current test setup. We need a better solution for this.

PR #324, opened 2022-11-17
fix(target): accept hex values for extended addresses

fkglr:
Currently no hex values are accepted in the target string for isotp extended address parameters.
@rumpelsepp is there any reason, not to use auto_int for all integer values?


rumpelsepp:
>  is there any reason, not to use auto_int for all integer values?

No. :)

PR #325, opened 2022-11-17
feat(scan-reset): added option to power-cycle on timeout

fkglr:
ECUReset can trigger shutdowns of ECUs which currently leads to TimeoutError, which in turns leads to termination of the scanner.
This commit, introduces the cli option `--power-cycle-on-timeout`. In case this option is set, the ECU power-cycles the ECU after such a timeout and tries to reconnect to the ECU afterwards.
Tested with isotp and connection-oriented protocols.

Co-authored by @VeroSec and @emedav

### DIFF
--- a/src/gallia/commands/scan/uds/reset.py
+++ b/src/gallia/commands/scan/uds/reset.py
@@ -146,7 +146,7 @@ class ResetScanner(UDSScanner):
                     else:
                         await self.ecu.wait_for_ecu()
 
-                except asyncio.TimeoutError:
+                except (asyncio.TimeoutError, ConnectionError) as e:
                     if args.power_cycle_on_timeout:
                         try:
                             self.logger.warning(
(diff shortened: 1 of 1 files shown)


rumpelsepp:
We could take the opportunity and refactor this thing. The `main()` is already way too long.

rumpelsepp:
Payload fuzzer potentially has similar problems; investigate if there is a shared solution. @emedav @VeroSec

rumpelsepp:
Reworked in #413.

fkglr:
As far as I can see, https://github.com/Fraunhofer-AISEC/gallia/pull/413 does not fully include all the cases which are handled here, so I don't think that this should be closed already.
Apart from that I appreciate the uncomplicated way this was handled there and hope that we can also fina a compromise bring this MR to master in the same fashion

rumpelsepp:
> does not fully include all the cases which are handled here

Which one is still missing?

rumpelsepp:
If this is still relevent, feel free to reopen.

PR #326, opened 2022-11-21
fix: handle PermissionError for lock files created by other users

ghost:
Co-Authored and tested by @fkglr.


rumpelsepp:
I do not support this change. What happens, when the user actually has no permission to write to the directory and the file does not exist? Try e.g. setting the path to the lockfile to `/foo.lock`.

rumpelsepp:
See this #327 proposal instead.

PR #327, opened 2022-11-21
fix: Do not fail when different owner of lockfile

rumpelsepp:
- Revert "fix: handle PermissionError for lock files created by other users (#326)"
- fix: Do not fail when different owner of lockfile

When the owner of the lockfile is different from the current user, `.touch()` fails. The lock can be created anyway. Only execute `.touch()` when the file does not exist.

### DIFF
--- a/src/gallia/command/base.py
+++ b/src/gallia/command/base.py
@@ -317,11 +317,7 @@ class BaseCommand(ABC):
         raise ValueError("base_dir or force_path must be different from None")
 
     def _aquire_flock(self, path: Path) -> None:
-        try:
-            # Raises 'PermissionError' when file was created by another user.
-            path.touch()
-        except PermissionError:
-            pass
+        path.touch()
         self._lock_file_fd = os.open(path, os.O_RDONLY)
         try:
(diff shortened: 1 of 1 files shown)


fkglr:
In my opinion, this more or less equivalent code would be more pythonic:

```python
        try:
            path.touch(exist_ok=False)
        except FileExistsError:
            pass
```

rumpelsepp:
A quick Google search says that exception based control flow is considered bad style. Exceptions should be used for error handling. In this case it doesn't handle anything.

edit: Also, in order to understand your code snippet, you have to look up the documentation. It is not clear what it is doing.

In contrast, this

``` python
if not file.exists():
    file.touch()
```

is easy to understand.

fkglr:
> A quick Google search says that exception based control flow is considered bad style. Exceptions should be used for error handling. In this case it doesn't handle anything.
> 
> edit: Also, in order to understand your code snippet, you have to look up the documentation. It is not clear what it is doing.
> 
> In contrast, this
> 
> ```python
> if not file.exists():
>     file.touch()
> ```
> 
> is easy to understand.

I'm fine with both.

PR #352, opened 2023-01-27
feat: Introduce new setting `gallia.power_cycle_sleep_after`

rumpelsepp:
The powercycle works like this:

1. power off
2. wait for `gallia.power_cycle_sleep` time
3. power on
4. wait for `gallia.power_cycle_sleep_after` time

Step 4 was hard coded to 2 secs before this change.


rumpelsepp:
Maybe define the powercycle wait stuff entirely better for gallia 2.0.

PR #370, opened 2023-02-24
feat(db): Add command_meta and settings from argparser as json

peckto:
This PR adds two columns to the `run_meta` db table:
1) `command_meta`: same as the one in `META.json`
2) `settings`: key -> value pair of settings retrieved from the argparser

For settings from the argparser, only the follwing trivial types are currently supported: `str`, `int`, `float`.
Other, complex data types would need a custom json serializer, which is out of scope for now.
Unsupported data types are skipped to prevent any confusion. 

The DB schema version is increased to `2.0`

### DIFF
--- a/src/gallia/db/handler.py
+++ b/src/gallia/db/handler.py
@@ -27,7 +27,7 @@ def bytes_repr(data: bytes) -> str:
     return bytes_repr_(data, False, None)
 
 
-schema_version = "2.0"
+schema_version = "2"
 
 DB_SCHEMA = f"""
 CREATE TABLE IF NOT EXISTS version (
--- a/src/gallia/db/handler.py
+++ b/src/gallia/db/handler.py
@@ -27,7 +27,7 @@ def bytes_repr(data: bytes) -> str:
     return bytes_repr_(data, False, None)
 
 
-schema_version = "1.0"
+schema_version = "2.0"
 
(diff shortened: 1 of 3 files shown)


src/gallia/db/handler.py
```diff
@@ -27,7 +27,7 @@ def bytes_repr(data: bytes) -> str:
     return bytes_repr_(data, False, None)
 
 
-schema_version = "1.0"
+schema_version = "1.1"
```
rumpelsepp: Let's use schema `2`. Major/Minor schemas do not make sense IMO.

src/gallia/utils.py
```diff
@@ -212,3 +213,15 @@ def lazy_import(name: str) -> ModuleType:
     sys.modules[name] = module
     loader.exec_module(module)
     return module
+
+
+def dump_args(args: Namespace) -> dict[str, str | int | float]:
+    settings = {}
+    for key, value in args.__dict__.items():
+        if isinstance(value, str):
```
rumpelsepp: use `match`

src/gallia/db/handler.py
peckto: fixed

src/gallia/utils.py
peckto: thanks, fixed

PR #413, opened 2023-07-13
feat(scan-reset): Power cycle ECU after timeout to recover state

peckto:
This PR extends the `scan-reset` scanner to power cycle the ECU in case of timeout (e.g. because we triggered a power-down reset level).
A power supply must be configured, otherwise this is ignored.

TODO:
- [x] Power cycle only if `--power-cycle` is provided
- [x] catch only `asyncio.TimeoutError` and `ConnectionError`


rumpelsepp:
There is already #325.

PR #433, opened 2023-09-15
feat(scan-services): Auto-iterate payload sizes for better discovery

ferdinandjarisch:
Extend the service scanner to automatically increase the payload size until no incorrectMessageLengthOrInvalidFormat error is encountered to succesfully judge whether the service is available.


rumpelsepp:
I think we could add a few words about this in the documentation as well:

https://gallia.readthedocs.io/en/latest/uds/scan_modes.html#service-scan

ferdinandjarisch:
> I think we could add a few words about this in the documentation as well:

done

PR #437, opened 2023-09-27
fix(dtc): Remove duplicate yet less capable primitive

ferdinandjarisch:


ferdinandjarisch:
https://github.com/Fraunhofer-AISEC/gallia/blob/master/src/gallia/commands/primitive/uds/dtc.py does the same thing but is way more capable.

rumpelsepp (approved):

PR #451, opened 2023-10-10
Improve DB handling

ferdinandjarisch:


rumpelsepp:
The DB is meant to be a Scanner feature; it provides very specific UDS features. Moving this into base is not a good fit.

ferdinandjarisch:
>it provides very specific UDS features

Those very specific UDS features are not utilized here.
Only the very generic DB features are utilized, that add generic META infos to the DB are.

Why doesn't it make sense to have this in the DB as well?

rumpelsepp:
> Why doesn't it make sense to have this in the DB as well?

We do not have a generic database class yet. From a datatype perspective, this mixes specific code with generic code and in my opinion this is a bad idea. You might not use the specific code in your patch, but the code is there. From my experience this will eventually grow and the code will end up more difficult to maintain.

So if you really want to have this, we need to separate the database code.

---

During writing this I looked up the code and I realized that what I am saying is actually not true. :dancers: :monkey:  The DB is part of `Scanner` and not `UDSScanner`. So, this in any case needs to be cleaned up for my argument to be valid. I think you could move it and we clean up later (tm). Opinions?

rumpelsepp (approved):

PR #454, opened 2023-10-24
chore(log): Slight improvement of log messages

ferdinandjarisch:


src/gallia/services/uds/server.py
```diff
@@ -387,7 +387,7 @@ async def setup(self) -> None:
         self.randomize()
 
         logger.notice(f"Initialized random UDS server with seed {self.seed}")
-        logger.info(
+        logger.notice(
```
rumpelsepp, line 390: I would rather say this is of level debug?

rumpelsepp (changes requested):

src/gallia/services/uds/server.py
ferdinandjarisch: How do you come to that conclusion?

The way I see it, the virtual ecu generates a random seed if none is given at start-up.
That seed then defines what services are available in which sessions (including available identifiers).

It totally makes sense to provide this information to the user, so that it is clear what the virtual ecu "is capable of", in this specific, randomly generated run.

rumpelsepp (approved):

PR #466, opened 2023-11-15
fix(ecu): Release the mutex while sleeping

rumpelsepp:
There was a massive performance regression, as
this mutex was hold while sleeping. Release the
mutex while sleeping to enable gallia doing stuff.

/cc @VeroSec


rumpelsepp:
@VeroSec Any new issues during your tests?

PR #467, opened 2023-11-20
cursed-hr: Add help view

fkglr:
Adds a help view to cursed-hr which can be opened by typing `?`.

### DIFF
--- a/src/cursed_hr/cursed_hr.py
+++ b/src/cursed_hr/cursed_hr.py
@@ -184,6 +184,18 @@ class CursedHR:
         self.configuration_index = 0
         self.use_prefix = use_prefix
 
+        self.priority_keys = {
+            "m": PenlogPriority.EMERGENCY,
+            "a": PenlogPriority.ALERT,
+            "c": PenlogPriority.CRITICAL,
+            "e": PenlogPriority.ERROR,
+            "w": PenlogPriority.WARNING,
+            "n": PenlogPriority.NOTICE,
+            "i": PenlogPriority.INFO,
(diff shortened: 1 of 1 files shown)


src/cursed_hr/cursed_hr.py
```diff
@@ -184,6 +184,18 @@ def __init__(
         self.configuration_index = 0
         self.use_prefix = use_prefix
 
+        self.priority_keys = {
+            "m": PenlogPriority.EMERGENCY,
```
rumpelsepp, line 188: maybe capital `E` instead?

rumpelsepp (approved):

src/cursed_hr/cursed_hr.py
fkglr: Personally, I would prefer to leave it this way. Besides, the keyboard mapping is not related to this pull request. This part of the code is only refactored to avoid code duplication / incosistencies between the help message and the real keyboard mapping.

ferdinandjarisch:
As said before, I would highly appreciate if a press of the `q` button while in the help view would only quit the help page, not the entire program.

Apart from that, I'm already super happy you worked on this! :)

fkglr:
> As said before, I would highly appreciate if a press of the `q` button while in the help view would only quit the help page, not the entire program.
> 
> Apart from that, I'm already super happy you worked on this! :)

This is now implemented in https://github.com/Fraunhofer-AISEC/gallia/commit/d1f16cd0d849a39faa8da37fe38682a5d9ccf1bf

PR #474, opened 2023-11-30
Small fixes here and there

ferdinandjarisch:


src/gallia/log.py
```diff
@@ -258,6 +258,9 @@ def setup_logging(
     logging.logThreads = False
     logging.logProcesses = False
 
+    # Silence python's root logger
```
ferdinandjarisch: This should probably only be done if len(logger_name) != 0

ferdinandjarisch:
@rumpelsepp What blocks this?

.github/workflows/tests.yml
```diff
@@ -73,6 +73,8 @@ jobs:
     - name: Add config
       run: |
         echo "[gallia]" > gallia.toml
+        echo "verbosity = 10" >> gallia.toml
```
rumpelsepp: verbosity of value 10 has no special meaning: edef9290411819fb7cc4d2bea237db58b73c9909. The default verbosity is INFO, then you could switch to DEBUG, that would be value `1` IIRC.

.github/workflows/tests.yml
```diff
@@ -73,6 +73,8 @@ jobs:
...
+        echo "no-volatile-info = true" >> gallia.toml
```
rumpelsepp: The volatile logging needs auto terminal detection; explicitely disabling this feature should not be needed IMO.

src/gallia/dumpcap.py
```diff
@@ -144,11 +144,15 @@ def _can_cmd(
         return args
 
     @staticmethod
-    async def _eth_cmd(target_ip: str) -> list[str]:
+    async def _eth_cmd(target_ip: str) -> list[str] | None:
+        if len(target_ip) == 0:
+            return None
```
rumpelsepp: I personally still consider a *failing and enabled* dumpcap a *critical error*, since expected logs will be missing in this case. So, the commit a40927de437a488e65105f477c03fc17070d3734 needs to be discussed: @peckto @fkglr. I personally would keep the current state as is and fail hard in this case. If you do not need dumpcap, it can be disabled (also via `gallia.toml`).

If there are valid reasons why `len(target_ip)` could be `0`, those should be handled by the caller, I guess.

src/gallia/log.py
```diff
@@ -258,6 +258,10 @@ def setup_logging(
...
+    if len(logger_name) != 0:
```
rumpelsepp: What's the rational of this change? f81561b533cf76f8d2119cc833d92956b2df10ff

src/gallia/services/uds/ecu.py
```diff
@@ -372,40 +372,19 @@ async def _tester_present_worker(self, interval: float) -> None:
         logger.debug("tester present worker started")
         while True:
             try:
-                async with self.mutex:
```
rumpelsepp, line 375: Did other people in our team test this commit 6ac8ad5131989ab42645264edb11e7c7e0a73b34 on ECUs from a different manufacturer? I would like to avoid  that changes in team A break setups in team B.

src/gallia/services/uds/ecu.py
```diff
@@ -416,13 +395,18 @@ async def start_cyclic_tester_present(self, interval: float) -> None:
...
     async def stop_cyclic_tester_present(self) -> None:
+        logger.debug("Stopping tester present worker")
         if self.tester_present_task is None:
             logger.warning(
                 "BUG: stop_cyclic_tester_present() called but no task running"
             )
             return
 
-        self.tester_present_task.cancel()
+        running = True
```
rumpelsepp: In my opinion, this commit fbc572333adb7ae131911f425ed755029925cd62 is not a proper fix for this not well understood problem of cancelling async tasks. It is still not clear why this problem actually occurs and if there are problems in the `gallia` codebase or elsewhere. So I would rather investigate the root cause of this. Maybe we should ask in the Python matrix channel or on the relevant mailing lists before merging this.

rumpelsepp (commented):
When you wrote your comment, this was still labeled as a draft, therefore I did not investigate further. I added my comments.

src/gallia/log.py
```diff
@@ -299,7 +303,7 @@ def add_stderr_log_handler(
 
 def add_zst_log_handler(
     logger_name: str, filepath: Path, file_log_level: Loglevel
-) -> None:
+) -> logging.Handler:
```
rumpelsepp, line 302: A major rework of the logging system is required with https://github.com/Fraunhofer-AISEC/gallia/discussions/479. This might eventually drop your proposed changes in the long term. If your suggested change is crucial for your work we could cherry pick them into a new pull request and merge them separetely.

.github/workflows/tests.yml
ferdinandjarisch: The numbers mason, what do they mean!

Let it be 1 then...

.github/workflows/tests.yml
ferdinandjarisch: Not sure what you mean?
I don't know how github jobs are set up and what should be done here...

In any case I see this as not related to this MR. Feel free to vote for dropping edef9290411819fb7cc4d2bea237db58b73c9909 again, I was just trying to debug the failing CI job...

src/gallia/dumpcap.py
ferdinandjarisch: I think you are misunderstanding.

The behavior you are describing is already on master: https://github.com/Fraunhofer-AISEC/gallia/blob/master/src/gallia/command/base.py#L552

This commit just handles incorrect dumpcap invocations, such as when a target like `unix-lines///tmp.sock` is given...
The main problem here is that the `else` branch is too broad: https://github.com/Fraunhofer-AISEC/gallia/blob/master/src/gallia/dumpcap.py#L64

src/gallia/log.py
ferdinandjarisch: Since we are not adjusting the root logger, it can be that other python modules that are using the `logging` module produce log output that is not intended, e.g. `numpy` or `matplotlib`.

This change disables these log entries.

src/gallia/services/uds/ecu.py
ferdinandjarisch: AFAIK this is tested on the vecu and ECUs from two manufacturers. But please be invited to test it further.

src/gallia/services/uds/ecu.py
ferdinandjarisch: Feel free to investigate!

Nonetheless I suggest we merge this to avoid the issue. As a compromise, we could add a `#FIXME` or the like...

src/gallia/log.py
ferdinandjarisch: I strongly disagree.

First, how does #479 affect "the logging system" at all?
Second, these are just passive changes that can easily be adjusted if needed.

rumpelsepp: As I mentioned, if you need those changes right now, please cherry pick them into a new pull request and I will merge them ASAP.

src/gallia/log.py
rumpelsepp: But then the change should be `if logger_name == "something"`, shouldn't it?

ferdinandjarisch: No, why should it?

You get the handle of the root logger by calling `logging.getLogger()` without arguments (or with an empty string).
Theoretically, our function `setup_logging` can also be used to configure the root logger directly instead of the default `gallia` logger.

This change is to ensure that the root logger is only silenced, if nobody is attempting to configure exactly the root logger, so basically `if logger_name != ""`, but linters told me checking for empty string is bad behavior.

src/gallia/log.py
ferdinandjarisch: I don't need them right now and can wait for the entire MR to be accepted, assuming this will also just be a few weeks max... :wink:

src/gallia/services/uds/ecu.py
rumpelsepp: You are free to investigate, too.

> I suggest we merge this to avoid the issue.

As far as I understand, this could create an infinite loop when the tasks decides to ignore cancelling, as it obviously does for some reason. Additionally, it could create an unexpected delay.

Perhaps we could avoid the cancelling problem entirely by using something like this:

``` python
# Mutexes or similar avoided for readability in this example.

# In tester present task
    # …
    while self.running:
        do_stuff()

# In main task
def stop_tester_present(self):
    self.running = False
```

If synchronization is required, there is [asyncio.Event](https://docs.python.org/3/library/asyncio-sync.html#asyncio.Event) which could be used to wait until the worker has been terminated.

src/gallia/services/uds/ecu.py
rumpelsepp: @peckto will do some test runs on his setup.

src/gallia/dumpcap.py
rumpelsepp: > The behavior you are describing is already on master: https://github.com/Fraunhofer-AISEC/gallia/blob/master/src/gallia/command/base.py#L552

You are right, I missed that and it is offtopic for this review here. However, I think this topic is still relevant and I will create an issue for this to discuss this further.

> This commit just handles incorrect dumpcap invocations, such as when a target like unix-lines///tmp.sock is given...

Yes, this is a special case in the function that handles arguments for ethernet setups. In the described case the `_eth_cmd` is called with wrong arguments. As mentioned, the main problem should be handled by the caller (i.e. the too broad else branch).

src/gallia/log.py
rumpelsepp: > No, why should it?

It is not clear why this feature is needed.

Further, there is no `numpy` or `matplotlib` in upstream `gallia`.

You mention:

> Theoretically, our function setup_logging can also be used to configure the root logger directly instead of the default gallia logger.

It is not clever to do special things, such as disabling the root logger in a function called `setup_logging()`. People relying on log messages from the root logger might spend hours of debugging and searching this hidden feature.

If I understand this correctly, the suggested code breaks the following code such that no messages are logged:

``` python
import logging

from gallia.log import setup_logging, get_logger

setup_logging()
logger = get_logger("")
logger.info("foo")

logging.info("foo")

logger = get_logger("root")
logger.info("foo")
```

Filtering log messages (by default) from extension modules is out of scope for upstream gallia.

> but linters told me checking for empty string is bad behavior.

This should be fine for linters:

``` python
if variable:
    foo()
```

src/gallia/services/uds/ecu.py
ferdinandjarisch: I did investigate, and this is what I came up with. If you want to dig deeper you of course can...

If you prefer the solution with a class variable that could work I guess, but keep in mind that it might be an overly complex workaround for a bug in ansyncio that might be fixed at some point?

Regarding your two points:

1. In theory this could be a infinite loop, yes. In practice it never went higher than 3 iterations, so I would take the chance. Keep in mind, Keep in mind that the task not being cancelled is a very rare/racy bug.
2. What do you mean by unexpected delay? For whom exactly and where would that be a problem?

src/gallia/dumpcap.py
ferdinandjarisch: >As mentioned, the main problem should be handled by the caller (i.e. the too broad else branch).

Both, no? So the else branch shouldn't be too broad (feel free to create MR), but the `_eth_cmd` can and should also do some basic sanity checking. The current approach with returning None matches e.g. [the behavior of `_can_cmd`](https://github.com/Fraunhofer-AISEC/gallia/blob/master/src/gallia/dumpcap.py#L135-L137).

ferdinandjarisch: And to state the (not so?) obvious: an empty string `target_ip` does not raise an error here: https://github.com/Fraunhofer-AISEC/gallia/blob/tp-shenanigangs/src/gallia/dumpcap.py#L151-L155

src/gallia/log.py
ferdinandjarisch: Well, what can I say apart from: "I did not come up with this as a programming exercise, but because we had unsolicited log messages".

If you have such a strong "wont fix" mindset here, I am fine with dropping f81561b533cf76f8d2119cc833d92956b2df10ff alltogether and just live with a few log messages that only exist on CLI. Let's not waste our time...

src/gallia/dumpcap.py
rumpelsepp: > Both, no? So the else branch shouldn't be too broad (feel free to create MR), but the _eth_cmd can and should also do some basic sanity checking. 

A sanity check of the arguments makes sence. However, the URI scheme is already checked in the caller and as mentioned this check is not complete. Let's fix this there.

> an empty string target_ip does not raise an error here

Yes, and I still do not see a valid reason for this. If there is no `target_ip` set, it makes no sense to call `_eth_cmd()`.

src/gallia/services/uds/ecu.py
rumpelsepp: Yeah, I am still not sure what is the best solution for this. I will have a look at this https://superfastpython.com/asyncio-task-cancellation-best-practices; maybe it helps.

Regarding points:

1. Yes, rare on one particular setup. Since nobody knows why and when this happens, it could behave differently on a different test setup.
2. When it takes more iterations to cancel the task, the scanner is blocked and it causes a delay that is not related to actual network traffic. This could lead to a lost UDS session, for instance. Not sure how realistic this scenario is.

I would like to test a few things on my machine first, in order to form a view on this problem.

src/gallia/dumpcap.py
ferdinandjarisch: will adapt once you merged #489

ferdinandjarisch: Adapted the behavior of `_eth_cmd` to match `_can_cmd` and make it return `None` on error.

Considering this still is the "default" due to `case _:` I think it remains a reasonable change.

src/gallia/services/uds/ecu.py
ferdinandjarisch: See fbcea4310c1184139a0fcaffce2a25a0e2169aba

Looks like a valid solution that avoids the `while True` approach.

I'm pretty sure the `asyncio.CancelledError` gets lost/consumed somewhere but no idea where and how.

src/gallia/dumpcap.py
VeroSec: Tested Branch. asyncio and teardown errors have been fixed. 
Only one error remains with python3.11:
`Exception ignored in: <function StreamWriter.__del__ at 0x7ff712588c20>
Traceback (most recent call last):
  File "/usr/lib/python3.11/asyncio/streams.py", line 395, in __del__
  File "/usr/lib/python3.11/asyncio/streams.py", line 343, in close
  File "/usr/lib/python3.11/asyncio/selector_events.py", line 860, in close
  File "/usr/lib/python3.11/asyncio/base_events.py", line 761, in call_soon
  File "/usr/lib/python3.11/asyncio/base_events.py", line 519, in _check_closed
RuntimeError: Event loop is closed`

ferdinandjarisch:
As mentioned by @VeroSec 

> Tested Branch. asyncio and teardown errors have been fixed.
> Only one error remains with python3.11:
> `Exception ignored in: <function StreamWriter.__del__ at 0x7ff712588c20> Traceback (most recent call last): File "/usr/lib/python3.11/asyncio/streams.py", line 395, in __del__ File "/usr/lib/python3.11/asyncio/streams.py", line 343, in close File "/usr/lib/python3.11/asyncio/selector_events.py", line 860, in close File "/usr/lib/python3.11/asyncio/base_events.py", line 761, in call_soon File "/usr/lib/python3.11/asyncio/base_events.py", line 519, in _check_closed RuntimeError: Event loop is closed`

This is an issue in cypthon and should be fixed in 3.11.7 onwards as far as I understood!

src/gallia/services/uds/ecu.py
ferdinandjarisch: I understand it was now tested by @VeroSec @fkglr @peckto  and myself.

src/gallia/services/uds/ecu.py
ferdinandjarisch: I understand you are OK with the changes in fbcea4310c1184139a0fcaffce2a25a0e2169aba ?

Care to resolve the conversation?

src/gallia/dumpcap.py
ferdinandjarisch: I understand you are OK with the changes in 37a368fbe883ccfa99e7c9ab49875aa56dba9eac ?

Care to resolve this conversation?

src/gallia/services/uds/server.py
```diff
@@ -871,6 +871,7 @@ async def handle_client(
                     await writer.drain()
             except Exception:
                 traceback.print_exc()
+                break
```
ferdinandjarisch, line 877: As far as I understood @fkglr prefers to specify certain exceptions that break the loop while others are ignored?

src/gallia/services/uds/server.py
```diff
@@ -899,6 +900,7 @@ async def run(self) -> None:
                     await transport.write(uds_response_raw)
             except Exception:
                 traceback.print_exc()
+                break
```
ferdinandjarisch, line 909: Currently this kills the ISOSTPServer since there is no recovery.

What exceptions do you expect that are recoverable, @fkglr ?

src/gallia/services/uds/server.py
ferdinandjarisch: `TCPUDSServerTransport`  accepts new connections anyhow, so we fail for all errors.

src/gallia/services/uds/server.py
ferdinandjarisch: `ISOTPUDSServerTransport` now dies after a single error, but let's see whether that becomes a problem and handle it once it occurs.

src/gallia/services/uds/ecu.py
rumpelsepp: > it was now tested …

Then let's move forward. ACK.

ferdinandjarisch:
@rumpelsepp rebased and ready to merge.

Does it need any more approval or can I press merge once jobs succeded?

PR #489, opened 2024-01-02
fix(dumpcap): Handle dumpcap invocation for unix transports

rumpelsepp:
Currently there is a flaw in the plugin interface. Plugins cannot
specify how dumpcap should be invoked to capture traffic. The default
is ethernet traffic. There are Unix transports available which are
neither CAN traffic nor ethernet traffic. In this case the wrong
method `_eth_cmd()` is called which cannot create a valid CLI for
dumpcap.

### DIFF
--- a/src/gallia/dumpcap.py
+++ b/src/gallia/dumpcap.py
@@ -18,7 +18,13 @@ from typing import cast
 from urllib.parse import urlparse
 
 from gallia.log import get_logger
-from gallia.transports import ISOTPTransport, RawCANTransport, TargetURI
+from gallia.transports import (
+    ISOTPTransport,
+    RawCANTransport,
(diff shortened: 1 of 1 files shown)


ferdinandjarisch:
Fixes the unix lines problem, but still has a pretty broad else branch...

Anyhow, looks good to me, but haven't tested it.

peckto (approved):

rumpelsepp:
> but still has a pretty broad else branch...

I added a point here: https://github.com/Fraunhofer-AISEC/gallia/discussions/479 under "Open for Discussion". We must default to ethernet for now, since our internal transports are ethernet. There is no API where transports could specify a command/something to capture traffic.

PR #496, opened 2024-01-11
chore(testing): Add basic unit test for Services Scanner

peckto:
This is a first draft of a unit test on scanner level.
The code uses `pytest` and the python inbuilt `unittest.mock` library.
Because of missing modularization, mocking the `argparse` library is necessary, 
which will be fixed in the future via a config container.
Currently, the transport module is mocked, but we could also create an own transport module for testing purpose.
As we currently have no defined scanner result, changes in the `scan-services` scanner are included for demonstration purposes.

To handle most of the default requests, the vECU is used.
To define test case specific ECU behavior, a callback `hook_request` is provided, which can overwrite the response from the vECU.

Three different ECU behaviors are implemented:
* PASS: forward response to the scanner (from the vECU or user defined)
* RAISE: raise defined exception instead of forwarding response
* TIMEOUT: simulate timeout during receive (realized as `TimeoutError` exception)

In the draft, three basic test cases for the scan-services scanner are included:

1. Basic test of Service Scan against vECU
2. Test TimeoutError on Service Scan
3. Test NegativeResponse on Service Scan

You can run the test cases via pytest:
```
pytest -k TestScanServices -s
```

### DIFF
--- a/src/gallia/commands/scan/uds/services.py
+++ b/src/gallia/commands/scan/uds/services.py
@@ -80,10 +80,10 @@ class ServicesScanner(UDSScanner):
     async def main(self, args: Namespace) -> None:
         self.result: list[tuple[int, int]] = []
         self.ecu.max_retry = 1
-        found: dict[int, dict[int, Any]] = {}
+        self.found: dict[int, dict[int, Any]] = {}
 
         if args.sessions is None:
-            found[0] = await self.perform_scan(args)
+            self.found[0] = await self.perform_scan(args)
         else:
             sessions = [
                 s
@@ -117,11 +117,11 @@ class ServicesScanner(UDSScanner):
 
                 logger.result(f"scanning in session {g_repr(session)}")
 
-                found[session] = await self.perform_scan(args, session)
+                self.found[session] = await self.perform_scan(args, session)
 
                 await self.ecu.leave_session(session)
 
-        for key, value in found.items():
+        for key, value in self.found.items():
             logger.result(f"findings in session 0x{key:02X}:")
             for sid, data in value.items():
                 self.result.append((key, sid))
@@ -163,6 +163,7 @@ class ServicesScanner(UDSScanner):
                     )
                 except asyncio.TimeoutError:
                     logger.info(f"{g_repr(sid)}: timeout")
+                    result[sid] = "TimeoutError"
                     continue
                 except MalformedResponse as e:
                     logger.warning(
--- a/tests/test_scan_services.py
+++ b/tests/test_scan_services.py
@@ -1,0 +1,215 @@
+# SPDX-FileCopyrightText: AISEC Pentesting Team
+#
+# SPDX-License-Identifier: Apache-2.0
+
+from collections.abc import Awaitable, Callable
+from dataclasses import dataclass
+from enum import Enum
+from typing import Any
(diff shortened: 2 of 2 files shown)


src/gallia/commands/scan/uds/services.py
```diff
@@ -80,10 +80,10 @@ def configure_parser(self) -> None:
     async def main(self, args: Namespace) -> None:
         self.result: list[tuple[int, int]] = []
         self.ecu.max_retry = 1
-        found: dict[int, dict[int, Any]] = {}
+        self.found: dict[int, dict[int, Any]] = {}
```
rumpelsepp, line 83: A unified interface for scanner results needs to be defined; this is not decided yet. So for these first tests, let's use your solution. We could adjust this later.

tests/test_scan_services.py
```diff
@@ -0,0 +1,215 @@
...
+from gallia.log import get_logger
+from gallia.services.uds import UDSErrorCodes, UDSIsoServices
+from gallia.services.uds.server import RandomUDSServer, UDSServerTransport
+from gallia.transports import TargetURI
+
+logger = get_logger("pytest")
+
+
+@dataclass
+class Args:
```
rumpelsepp, line 22: For now ok; in gallia 2.0 this will be replaced by config containers.

tests/test_scan_services.py
```diff
@@ -0,0 +1,215 @@
...
+
+    @pytest.mark.asyncio
+    async def test_2(self, setup: None, teardown: None) -> None:
+        """Test TimeoutError on Service Scan.
+        On SID 0x27 a TimeoutError is simulated.
+        The Scanner should return this in the result.
+        """
+        assert self.scanner is not None
+
+        async def handle(
```
rumpelsepp, line 174: As discussed via phone: This will be replaced via a table with TypedDicts, as suggested by russ cox: https://research.swtch.com/testing. Especially his tips 1, 5, 8, and 11 are relevant here.

Your draft from #495 could be used for a starting point.

tests/test_scan_services.py
```diff
@@ -0,0 +1,215 @@
...
+            return Action.PASS, res
+
+        self.hook_request = handle
+
+        await self.scanner.main(self.args)  # type: ignore
+
+        assert self.scanner.found[1][0x27] == TimeoutError.__name__
+
+    @pytest.mark.asyncio
+    async def test_3(self, setup: None, teardown: None) -> None:
```
rumpelsepp, line 188: Find descriptive test case names

tests/test_scan_services.py
```diff
@@ -0,0 +1,215 @@
...
+        con.close.side_effect = self.mock_close
+        con.write.side_effect = self.mock_write
+        con.read.side_effect = self.mock_read
+        con.reconnect.side_effect = self.mock_reconnect
+        con.request_unsafe.side_effect = self.mock_request_unsafe
+        return con
+
+    @pytest.mark.asyncio
+    @pytest.fixture
+    @patch("gallia.transports.unix.UnixTransport.connect")
```
rumpelsepp, line 134: Let's create a separate vecu transport for testing.

rumpelsepp:
Superseeded by #803

PR #522, opened 2024-04-02
dump-seeds: remove arbitrary default duration

fkglr:
Remove the arbitrary default of 12 hours, by default it should run until halted.
While on could argue, that this bears the risk of filling up the disk with random numbers, there still is not real reason for this 12h value, because in practice this could take hours to months depending on the output rate of the device under test and the remaining capacity of the system.


ferdinandjarisch (approved):
Looks sane and reasonable to me!

PR #523, opened 2024-04-05
fix(power-cycle): Make all occurrences of power cycles respect the --power-cycle-sleep parameter.

ghost:
Using `--power-cycle-sleep`, scanners can be told to sleep for a certain amount of time between power-off and power-on of the test-setup. However, this parameter was only actually used at one of multiple occurrences of these events in the code so far.

### DIFF
--- a/src/gallia/services/uds/ecu.py
+++ b/src/gallia/services/uds/ecu.py
@@ -209,27 +209,31 @@ class ECU(UDSClient):
         self.state.reset()
         return True
 
-    async def leave_session(self, level: int, config: UDSRequestConfig | None = None) -> bool:
+    async def leave_session(
+        self,
+        level: int,
+        config: UDSRequestConfig | None = None,
+        sleep: int | None = None,
+    ) -> bool:
         """leave_session() is a hook which can be called explicitly by a
         scanner when a session is to be disabled. Use this hook if resetting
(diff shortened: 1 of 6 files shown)


src/gallia/services/uds/ecu.py
```diff
@@ -209,27 +209,29 @@ async def callback() -> None:
...
         if isinstance(resp, service.NegativeResponse):
-            await self.power_cycle()
+            await self.power_cycle(**power_cycle_args)
             await self.reconnect()
         await self.wait_for_ecu()
 
         resp = await self.set_session(0x01, config=config)
         if isinstance(resp, service.NegativeResponse):
-            await self.power_cycle()
+            await self.power_cycle(**power_cycle_args)
```
rumpelsepp: Why not just `self.power_cycle(sleep)`?

ghost: To allow for using the default sleep time defined in `power_cycle()` without repeating its definition in `leave_session()`. The latter might be easily overseen, should the definition in `power_cycle()` be updated in the future.

rumpelsepp: When you use `**power_cycle_args` it does not get catched by `mypy`. When you use plain `sleep` as an argument it is all statically typed.

PR #525, opened 2024-04-19
docs: Improving documentation

mich41v4294:
PR Focused on improving documentation to ease the learning curve for newcomers, since a lot of documentation was lacking.

Edited rendered documentation available at [ReadTheDocs](https://mich-gallia-fork.readthedocs.io/).

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -4,7 +4,7 @@ version: 2
 build:
   os: "ubuntu-22.04"
   tools:
-    python: "3.10"
+    python: "3.11"
   commands:
     - curl -sSL https://install.python-poetry.org | python3 -
     - $HOME/.local/bin/poetry install
--- a/docs/api.md
+++ b/docs/api.md
@@ -38,14 +38,6 @@ SPDX-License-Identifier: CC0-1.0
     :show-inheritance:
 ```
 
-## gallia.commands
-
-```{eval-rst}
-.. automodule:: gallia.commands
-    :members:
-    :show-inheritance:
-```
-
 ## gallia.transports
 
 All available transports are documented in {doc}`../transports`.
--- a/docs/api.md
+++ b/docs/api.md
@@ -38,10 +38,10 @@ SPDX-License-Identifier: CC0-1.0
     :show-inheritance:
 ```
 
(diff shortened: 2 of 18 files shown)


rumpelsepp:
Thank you very much for your effort and your pull request! Very much appreciated!

There are a few remarks:

* We avoid documenting, e.g., command line flags in the documentation text. The reason is that it gets outdated very quickly when the corresponding code is changed. We have limited resources and we try to focus as much as possible. Thus, we document everything (if possible) in the code directly, e.g. in the help text of the argparser for the relevant `--help` pages.
* The documentation page ["scan_modes"](https://fraunhofer-aisec.github.io/gallia/uds/scan_modes.html) is a high level description of how the scanning works. It is very important for us that this page does not turn into a user manual, as this site is often used as a reference in reports etc. The focus here is "how does it work?" and not "how is it used?".

I find your explanations how particular scanners work useful. However, the syncronization problem code <-> documentation still applies. Looking forward, I propose the following:

1. *Move* docstrings that describe *how* a scanner works to  ["scan_modes"](https://fraunhofer-aisec.github.io/gallia/uds/scan_modes.html).
2. Remove duplications from the docstrings, e.g. documentation of command line arguments or options. Update the help strings of the relevant argparser instead. gallia 2.0 will use pydantic containers where the documentation of options, command line arguments or api settings can be placed at a single location in code. We have a tool that will migrate the code with that documentation to the new code architecture.
3. Remove `gallia.commands` from the API docs. It is not considered public API and in any case it will change with gallia 2.0. The proposed docstrings can be moved to a (new) page in the sphix documentation.

Thank you for your understanding. I took some time and thought about how to write this answer. I hope it generates only a decent amount of additional work for you!

mich41v4294:
I have removed duplicate information from the docstrings, made `scan_modes.md` focus more on the actual technical workings of the scans so they could be referenced in reports and specific scans understood by third parties without the knowledge of the tool and removed `gallia.commands` from the API.

I would also like to ask for confirmation, Gallia ISOTP Discover does not support 29bit CAN IDs now at all? Wanted to note that in the ISOTP chapter of `scan_modes.md` but couldn't get it working.

PR #527, opened 2024-04-22
feat: isotp discovery - try to discover servers with both padding options by default

mich41v4294:
Added functionality for the isotp UDS discovery to scan twice by default - once without padding and once with. Abstracted the scanning loop in the process.

### DIFF
--- a/src/gallia/commands/discover/uds/isotp.py
+++ b/src/gallia/commands/discover/uds/isotp.py
@@ -46,7 +46,7 @@ class IsotpDiscoverer(UDSDiscoveryScanner):
             "--padding",
             type=string_dict,
             default="auto",
-            help="set isotp padding (default: auto - try both)",
+            help="set isotp padding (default: auto: try both)",
         )
         self.parser.add_argument(
             "--pdu",
--- a/src/gallia/commands/discover/uds/isotp.py
+++ b/src/gallia/commands/discover/uds/isotp.py
@@ -11,7 +11,7 @@ from gallia.log import get_logger
 from gallia.services.uds import NegativeResponse, UDSClient, UDSRequest
 from gallia.services.uds.core.utils import g_repr
 from gallia.transports import ISOTPTransport, RawCANTransport, TargetURI
-from gallia.utils import auto_int, can_id_repr, write_target_list
+from gallia.utils import auto_int, can_id_repr, write_target_list, string_dict
 
 logger = get_logger("gallia.discover.isotp")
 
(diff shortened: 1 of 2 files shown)


mich41v4294:
I your guys opinions on this: how should be duplicity handled? Is it important to know that the UDS server responds to both padded and non-padded messages? shall we log both or somehow remove the second hit? Unsure how to approach this. @rumpelsepp would you be able to shed more light on this maybe?

As of the [3dd1b32](https://github.com/Fraunhofer-AISEC/gallia/pull/527/commits/3dd1b3266b49fc4a6dc0f570298e0c5fec1a36c3) commit, the code all works good, but logs the targets twice.

rumpelsepp:
Do you have a specific use case for scanning the padding? In the past we repeatedly failed with reliably discovering the settings for a transport. Consequently, we now consider this as a required prior knowledge for this scanner. There is more prior knowledge that is comparable to this one, e.g., CAN Rate, CAN-FD vs classic CAN, …. Often these settings are known by documentation or by tinkering…

Additionally, the search space will be increased by one byte for the full range; thus increasing the runtime of this scanner significantly.

mich41v4294:
> Do you have a specific use case for scanning the padding?

Yes, some ECUs only respond to one padding configuration and this made me confused when using gallia initially, because the tool I was used to previously did scan with both padding and no-padding automatically, sometimes there's even ECUs with different padding behavior in a single vehicle.

> we now consider this as a required prior knowledge for this scanner

I'd argue that it's not the same level of prior knowledge, since the CAN configuration would be defined in the usual documentation (dbc) and it's something you expect to be different on every VUT. Meanwhile you could assume UDS is a common standard and having an unknown here is a bigger problem than in CAN configuration.

CAN settings can also be automatically discovered (on interfaces that support `berr-reporting`, tested on a [PiCAN FD Duo](https://www.skpang.co.uk/products/pican-fd-duo-board-with-real-time-clock-for-raspberry-pi-3)) and along with the padding scanning could serve as an extension of the advanced automation features in `Gallia 2.0`.

> increasing the runtime of this scanner significantly

While this is correct, the PR includes a CLI argument to not perform the scanning twice, but I'm proposing to have automatic as a default, since if you already know if your UDS server needs padding/no-padding, you know that you should look for it as an CLI arg.

rumpelsepp:
This has been addressed in: https://github.com/Fraunhofer-AISEC/gallia/pull/807; especially in: https://github.com/Fraunhofer-AISEC/gallia/pull/807/changes/77ca2ada3c90174344f2af3fdfe4f876c0fa1bc1

Thank you for your pull request!

PR #530, opened 2024-04-24
feat: added security access subfunction scanner

mich41v4294:
Added a scan mode to enumerate supported SecurityAccess (0x27) subfunctions/levels in supported sessions.

### DIFF
--- a/docs/uds/scan_modes.md
+++ b/docs/uds/scan_modes.md
@@ -147,6 +147,76 @@ For discovering available subFunctions the following error codes indicate the su
 
 Each identifier or subFunction which responds with a different error code is considered available.
 
+## Security Access Level Scan
+
+This type of scan searches for available security access levels (SAs) within a UDS server on specified sessions. Security access levels are used to restrict access to certain UDS services and subfunctions. By identifying the available SAs, an attacker might gain insights into the ECU's security mechanisms and potentially exploit them to elevate privileges.
+
+The `gallia` tool offers a class `SALevelScanner` to perform this security access level scan.
+
+### Usage
+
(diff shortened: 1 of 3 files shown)


mich41v4294:
should've read the documentation earlier, found out the `identifiers` scan should be used for this. feel free to merge/close as per your discretion. my thought was this specific scan would enable possible future automation expansion, but that could be also done with the universal `identifiers` scan. Only benefit here might be more specific response/data handling.

peckto:
Hi @mich41v4294 
thanks for your contribution!
Yea, you can use the identifier scanner to iterate the sa level.
I saw, that you also iterate the payload length to request the seed.
Is there a specific reason why you are doing this? 
Do you get different results on different payload lengths?

mich41v4294:
> Do you get different results on different payload lengths?

Yes, that's correct, some ECUs might expect additional information with the SA request and could respond with `IncorrectLength`. While that response itself is enough to assess the existence of that specific SA level, it is not enough to specify the correct message format (eg. for subsequent seed gathering).

peckto:
Ok, I see.
Maybe we can split this up?
We like to have simple scanners, which are focused on one thing, in case of the identifiers scanner, finding IDs which are available. A second scanner could then take those IDs and try to find accepted payload lengths. A third could then use the payload length to do fuzzing (where we have a prototype). 
Decoupling adds flexibility to run (or re-run) only selected parts and allows combining and reusing the scanners as needed.

rumpelsepp:
Thank you for your pull request! Since this is already covered by the `identifiers` scanner and this is now outdated due to our API remake for gallia 2.0, I am closing this.

PR #539, opened 2024-05-13
fix(uds-scan-memory): leave session after scan.

ghost:
`gallia uds scan identifiers` leaves each session before moving on to the next or terminating. `gallia uds scan memory` does not do so. This pull request aligns the behavior of the latter with the identifier scanner.


src/gallia/commands/scan/uds/memory.py
```diff
@@ -63,6 +63,10 @@ async def main(self, args: Namespace) -> None:
         for i in range(5):
             await self.scan_memory_address(args, i)
 
+        logger.result(f"Scan in session {g_repr(args.session)} is complete!")
```
rumpelsepp: Make this level info and I'll merge it.

PR #544, opened 2024-06-19
fix(cursed-hr): fallback to source dir if tmp dir not working

fkglr:
In case the tmp directory is too small (e.g. if in small RAM), decompressing large files fails.
This PR adds a fallback to creating a temporary file in the directory of the input file.

### DIFF
--- a/src/cursed_hr/cursed_hr.py
+++ b/src/cursed_hr/cursed_hr.py
@@ -23,6 +23,7 @@ from math import ceil
 from pathlib import Path
 from typing import Any, BinaryIO
 
+import platformdirs
 import zstandard as zstd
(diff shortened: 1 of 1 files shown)


rumpelsepp:
`gallia` conforms to the [xdg standard](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html). I would suggest to use `$XDG_CACHE_HOME/gallia/filename` for this. `$XDG_CACHE_HOME` falls back to `~/.cache`.

[This library](https://pypi.org/project/platformdirs/) is already a dependency. You can just use their `user_cache_dir`.

edit: `gallia.log.PenlogReader` might benefit from this change, too.

PR #556, opened 2024-07-17
chore: Cleanup logger creation

rumpelsepp:
Run:

        $ rg -l 'get_logger\(".*"\)' | xargs sed -iE 's|get_logger(".*")|get_logger(__file__)|'

`get_logger(__file__)` does what you expect without typos.


ferdinandjarisch:
Hm, I seem to remember that in my initial draft of the logging-rework I already took this approach but you/Tobi somehow wanted the hardcoded values... :thinking:

Anyhow, haven't tested this, but content-wise it looks fine to me!

PR #558, opened 2024-07-17
fix(log): Fix wrong timezone in logfiles

### LINKED ISSUE #557: Logged time is UTC without time zone information/adjustments
Currently the logs only store plain UTC timestamps for each entry.
This can lead to confusion when viewing them, as they appear to be "out of sync" with the machine on which they are taken.

This behavior might have been introduced in this commit: https://github.com/Fraunhofer-AISEC/gallia/commit/939f26764cbf1e9fc274d912f85d31e0b3235cba

rumpelsepp:
Since 939f26764cbf1e9fc274d912f85d31e0b3235cba the logfiles contained
the UTC timezone which is wrong. This patch changes the code such that
the timezone is set correctly to the **local** timezone of the current
platform.

### DIFF
--- a/src/gallia/log.py
+++ b/src/gallia/log.py
@@ -15,6 +15,7 @@ import shutil
 import socket
 import sys
 import tempfile
+import time
 import traceback
 from collections.abc import Iterator
 from dataclasses import dataclass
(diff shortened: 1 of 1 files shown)


rumpelsepp:
tested by me and @peckto

PR #559, opened 2024-07-18
chore: Migrate to new context manager asyncio.timeout()

rumpelsepp:


ferdinandjarisch:
What exactly is the added benefit or the fixed bug of this change?

src/gallia/transports/doip.py
```diff
@@ -661,17 +661,17 @@ async def write_request_raw(self, hdr: GenericHeader, payload: DoIPOutData) -> N
...
                             / 1000,
-                        )
+                        ):
+                            await self._read_ack(payload.UserData)
                     case RoutingActivationRequest():
-                        await asyncio.wait_for(
-                            self._read_routing_activation_response(),
-                            TimingAndCommunicationParameters.RoutingActivationResponseTimeout
+                        async with asyncio.timeout(
+                            TimingAndCommunicationParameters.DiagnosticMessageMessageAckTimeout
```
ferdinandjarisch: We should avoid introducing new bugs, especially with such seemingly straightforward changes...

ferdinandjarisch (changes requested):

rumpelsepp:
> What exactly is the added benefit or the fixed bug of this change?

Code consistency. This is the recommended way of doing async timeouts nowadays: https://docs.python.org/3/library/asyncio-task.html#timeouts.

I started to use this new API in #555 which would introduce a new way of doing things. For the sake of consistency I did this migration. IMO it is important to stay syncronized with how the community does things to keep the codebase evolving.

PR #562, opened 2024-07-24
fix(logger): use name instead of file for logger base

peckto:
Fixup for #556
Unfortunately, `__file__` does not work, because it returns the full path to the source file (eg. `/path-to-venv/gallia/...`) and the logger name is expected to be a dot-separated tree with `gallia` as base (because we filter on that).

I propose to use the `__name__` variable for the logger name, it starts with `gallia.`.


ferdinandjarisch (approved):
As said, I didn't test #556 but yeah, `__name__` is what I had in mind, as this is what python's logging docs are suggesting to use!

PR #563, opened 2024-07-26
chore(tests): Raise pytest Warnings as Errors instead

ferdinandjarisch:
Since we are doing this for our plugins, it makes sense to also check for it here.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -41,7 +41,7 @@ jobs:
         poetry install
     - name: Run pytest
       run: |
-        poetry run python -m pytest -v tests -W error::UserWarning
+        poetry run make test
 
   vecu:
     strategy:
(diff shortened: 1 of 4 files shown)


rumpelsepp:
May you try if this works for you instead? If we use the setting in the `pyproject.toml` then any IDE, LSP, or whatever tooling people are using has the same settings when running the tests.

https://docs.pytest.org/en/stable/how-to/capture-warnings.html#controlling-warnings

```
# pyproject.toml
[tool.pytest.ini_options]
filterwarnings = [
    "error",
    "ignore::UserWarning",
    # note the use of single quote below to denote "raw" strings in TOML
    'ignore:function ham\(\) is deprecated:DeprecationWarning',
]
```

ferdinandjarisch:
> May you try if this works for you instead? If we use the setting in the `pyproject.toml` then any IDE, LSP, or whatever tooling people are using has the same settings when running the tests.

The example that you posted obviously does not work, but using `pyproject.toml` in general does.
Added a fixup commit with this approach; does that match what you had in mind?

rumpelsepp (approved):

PR #572, opened 2024-08-05
feat: SecurityAccess key length scanner

ghost:
Adds a very basic scanner to determine the key length expected by SecurityAccess.


rumpelsepp:
@emedav should this be in gallia 2.0?

ghost:
See #684.

PR #573, opened 2024-08-05
feat(sa_dump_seeds): Add sleep parameter and output seed length

ghost:
Adds a sleep parameter as an additional way to attempt fooling brute-force protections. Also outputs the received key length.

### DIFF
--- a/src/gallia/commands/scan/uds/sa_dump_seeds.py
+++ b/src/gallia/commands/scan/uds/sa_dump_seeds.py
@@ -220,7 +220,7 @@ class SASeedsDumper(UDSScanner):
 
             if args.sleep > 0:
                 logger.info(f"Sleeping for {args.sleep} seconds between seed requests...")
-                time.sleep(args.sleep)
+                await asyncio.sleep(args.sleep)
 
         await file.close()
         self.log_size(seeds_file, time.time() - start_time)
--- a/src/gallia/commands/scan/uds/sa_dump_seeds.py
+++ b/src/gallia/commands/scan/uds/sa_dump_seeds.py
@@ -84,6 +84,13 @@ class SASeedsDumper(UDSScanner):
             default=b"",
             help="Append an optional data record to each seed request",
         )
+        self.parser.add_argument(
+            "--sleep",
+            default=0,
+            type=float,
+            metavar="FLOAT",
+            help="Attempt to fool brute force protection by sleeping for N seconds between seed requests.",
+        )
 
     async def request_seed(self, level: int, data: bytes) -> bytes | None:
         resp = await self.ecu.security_access_request_seed(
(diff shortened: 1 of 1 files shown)


src/gallia/commands/scan/uds/sa_dump_seeds.py
```diff
@@ -209,6 +218,10 @@ async def main(self, args: Namespace) -> None:
                 # Re-enter session. Checking/logging will be done at the beginning of next iteration
                 await self.ecu.set_session(session)
 
+            if args.sleep > 0:
+                logger.info(f"Sleeping for {args.sleep} seconds between seed requests...")
+                time.sleep(args.sleep)
```
rumpelsepp:
```suggestion
                await asyncio.sleep(args.sleep)
```

src/gallia/commands/scan/uds/sa_dump_seeds.py
```diff
@@ -209,6 +218,10 @@ async def main(self, args: Namespace) -> None:
                 # Re-enter session. Checking/logging will be done at the beginning of next iteration
                 await self.ecu.set_session(session)
 
+            if args.sleep > 0:
+                logger.info(f"Sleeping for {args.sleep} seconds between seed requests...")
```
rumpelsepp:
```suggestion
                logger.info(f"Sleeping for {args.sleep} seconds between seed requests…")
```

src/gallia/commands/scan/uds/sa_dump_seeds.py
```diff
@@ -209,6 +219,10 @@ async def main(self, args: Namespace) -> None:
                 # Re-enter session. Checking/logging will be done at the beginning of next iteration
                 await self.ecu.set_session(session)
 
+            if args.sleep > 0:
```
rumpelsepp: So above and then:

```suggestion
            if args.sleep is not None:
```

src/gallia/commands/scan/uds/sa_dump_seeds.py
```diff
@@ -84,6 +85,13 @@ def configure_parser(self) -> None:
             default=b"",
             help="Append an optional data record to each seed request",
         )
+        self.parser.add_argument(
+            "--sleep",
+            default=0,
```
rumpelsepp: nitpick: Remove this, then you can do a `not None` check below. Would also work better for gallia 2.0 with stricter typing.

```suggestion
```

src/gallia/commands/scan/uds/sa_dump_seeds.py
rumpelsepp: The most elegant solution might be this one: Leave the `default=0` thing and just remove the `if`. I think `await asyncio.sleep(0)` is a noop. Choose your favorite. :)

PR #575, opened 2024-08-07
refactor(ECU.connect): move definition to UDSClient and call it as part of UDSClient.reconnect

ghost:


rumpelsepp:
I think this method is not used any more: https://github.com/search?q=repo%3AFraunhofer-AISEC%2Fgallia%20.connect()&type=code

We had this in a historic version IIRC. Is there code that should be in `.connect()`? If not, I would just remove this unused method instead.

ghost:
> I think this method is not used any more: https://github.com/search?q=repo%3AFraunhofer-AISEC%2Fgallia%20.connect()&type=code
> 
> We had this in a historic version IIRC. Is there code that should be in `.connect()`? If not, I would just remove this unused method instead.

It is not currently used by any code in this repository, but with the change I proposed it becomes a nice hook for vendor-specific setup code required in some of our proprietary plug-ins.

rumpelsepp:
I see. So, `.connect()` in those cases triggers the `.connect()` of the transport underneath, does it?

rumpelsepp (approved):

PR #579, opened 2024-08-21
feat: Expose the response pending timeout to CLI and config

rumpelsepp:
When using FlexRay the ISO-TP layer might be too slow for the UDS stack.
In our setup the UDS stack timeouted before the ISO-TP code was ready
with resolving a response pending.

This patch exposes the timeout value in order to be able to configure
this problem away.


rumpelsepp:
This does not apply to the codebase anymore, since this code is for the obsolete argparse based config. We migrated to pydantic in the meantime.

PR #589, opened 2024-09-02
chore: Remove more-itertools dependency

rumpelsepp:
Python >= 3.12 includes this iterator in the stdlib. Let's wait until
Python 3.13 is released before merging.


rumpelsepp:
Release date of Python 3.13 is 2024-10-01. I suspect that gallia 2.0 comes after this date. Thus I added this to gallia 2.0.

PR #593, opened 2024-09-19
feat: Add additional logging tags for protocol information

rumpelsepp:
This MR adds the following; currently DRAFT for experiments:

* standardize logging tags:
  * `io={TYPE}`; `TYPE` is either `read` or `write`
  * `proto={PROTO}`; `PROTO` is one of the supported protocols in gallia, e.g. `uds` or `doip`.
  * `encoding={ENCODING}`; optional; `ENCODING` could be `hex` or `base64`
* add a dissector API
  * a dissector could be used in the `hr` commands to interpret the tagged messages

The motivation of this is, we have currently a ["works somehow" dissector](https://github.com/Fraunhofer-AISEC/gallia/blob/9ac497585ded867ddf1de8eafcbf0f598b2bf2f9/src/cursed_hr/cursed_hr.py#L626-L642) in `cursed_hr` for UDS. This dissector uses a best effort heuristic to detect UDS messages. However, this is not useful for messages of other types.

gallia recently gained FlexRay support and a standardized dissection API in our `hr` commands would help everybody. For instance, for FlexRay there is already [an own tool](https://github.com/Fraunhofer-AISEC/gallia/blob/9ac497585ded867ddf1de8eafcbf0f598b2bf2f9/src/gallia/commands/script/flexray.py#L43-L54) to help debugging.

So, this MR is an attempt to standardize dissection in our toolchain by adding metadata to our logfiles and an extendable API for `hr`.

Open Questions:

* API for dissector: How should protocol metadata be handled? e.g. FlexRay Slot ID or CAN IDs?

### DIFF
--- a/src/gallia/dissect/__init__.py
+++ b/src/gallia/dissect/__init__.py
@@ -1,0 +1,48 @@
+# SPDX-FileCopyrightText: AISEC Pentesting Team
+#
+# SPDX-License-Identifier: Apache-2.0
+
+from abc import ABC, abstractmethod
+from dataclasses import dataclass
+from typing import Any, Literal, NotRequired, TypeAlias, TypedDict
+
+from gallia.services.uds.core.service import UDSRequest, UDSResponse
+
+
+@dataclass
+class Field:
+    name: str
+    raw_data: bytes
+    dissected_data: str
+
+
+class BaseDissector(ABC):
+    PROTO: str = ""
+
+    def __init_subclass__(
+        cls,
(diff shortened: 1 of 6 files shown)


rumpelsepp:
Superseded by #893: the protocol is stored in a dedicated (unstable) field `_proto` instead of `proto=` tags, the direction stays in the existing `read`/`write` tags, and the dissector registry lives in `gallia.dissect`.

PR #598, opened 2024-09-24
chore: Remove tabulate dependency

rumpelsepp:
This is only used in one location for a really simple table. Let's get
rid of the dependency.

**Needs testing**.

see also: https://github.com/Fraunhofer-AISEC/gallia/pull/582, https://github.com/Fraunhofer-AISEC/gallia/pull/590, https://github.com/Fraunhofer-AISEC/gallia/pull/591, https://github.com/Fraunhofer-AISEC/gallia/pull/595, https://github.com/Fraunhofer-AISEC/gallia/pull/596 #597


rumpelsepp:
I think this is not worth it. Let's keep this dep for now. Maybe reconsider later.

PR #613, opened 2024-10-24
feat: Add support for RND power supplies

rumpelsepp:
A very limited port of this:
https://github.com/rumpelsepp/opennetzteil/blob/master/devices/rnd/rnd320.go

However, these devices are so buggy that only switch on/off works
reliably … (in most cases).


rumpelsepp:
I merge this, since HTTP support has been removed in #597. Once anybody needs the RND power supplies there is at least this starting point here available.

PR #635, opened 2024-12-03
DoIP fixes and grand speed-up in discovery

ferdinandjarisch:


src/gallia/commands/discover/doip.py
```diff
@@ -2,6 +2,8 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
+from __future__ import annotations
```
rumpelsepp: Is [deprecated](https://docs.python.org/3.14/whatsnew/3.14.html#from-future-import-annotations) in cpython. Does it work when this line is removed?

src/gallia/commands/discover/doip.py
```diff
@@ -545,3 +543,56 @@ async def run_udp_discovery(self) -> list[tuple[str, int]]:
             )
 
         return found
+
+
+class FastDoIPConnection(DoIPConnection):
+    # This code is pretty much copied from the original except for the special treatment of DiagnosticMessages which
```
rumpelsepp: Copy + pasted code is a bad idea for long term maintenance. What is the limitation to include this functionality into `DoIPConnection` directly?

rumpelsepp:
Looks good. However, I would motivate to avoid copy+pasted code and refactor this part.

src/gallia/commands/discover/doip.py
ferdinandjarisch: Obviously, introducing a second queue specifically for DiagnosticMessages completely changes the behavior of the `DoIPConnection`. Since I do not see a use case for this apart from the discoverer I thus decided to place it here instead.

If you think it makes more sense to incorporate it in the general `DoIPConnection`, I can easily do so with a flag to separate the queues in the `__init__` function?

rumpelsepp: I think this would be a better solution. Then you have one entry point (= the DoipConnection class) that could be configured to behave in a certain way.

src/gallia/commands/discover/doip.py
ferdinandjarisch: Due to the changes with `DoIPConnection` it is no longer needed anyhow.

src/gallia/commands/discover/doip.py
ferdinandjarisch: I placed fixme commits on top. Please review whether this is to your liking!

PR #643, opened 2024-12-09
Remove volatile-info in favor of a bottom progress bar (cf. apt install)

### LINKED ISSUE #168: Implement progress bar
Use this: 

* https://pypi.org/project/tqdm
* https://rich.readthedocs.io/en/latest/progress.html

rumpelsepp:
Remove the workaround `--volatile-info` and create a progress bar, which is inspired by `apt install`. The progress bar is only UI and only visible in the terminal, thus is not logged.

https://github.com/user-attachments/assets/b97a8b2c-ed90-432b-9392-7135277431ce

### DIFF
--- a/src/gallia/cli/gallia.py
+++ b/src/gallia/cli/gallia.py
@@ -185,7 +185,6 @@ def parse_and_run(
     if setup_log:
         setup_logging(
             level=get_log_level(config.verbose),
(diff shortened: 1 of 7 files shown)


rumpelsepp:
As discussed with @ferdinandjarisch, this does not solve the problem of breaking log output of, e.g., pytest or similar. Closing this for now.

PR #644, opened 2024-12-09
chore: Introduce pip support and make uv optional

ferdinandjarisch:
The current implementation of the pyproject.toml does not allow to install all requirements for dev environments to be installed with pip. While PEP 735 is accepted since almost two months, pip does not support it yet.

Instead, we rely on PEP 621 and by using optional-dependencies we can for now support pip by allowing to install the dev dependencies by specifying extras, such as

pip install .[dev]

This can be reverted once https://github.com/pypa/pip/issues/12963 is resolved.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,38 +16,38 @@ default:
 
 .PHONY: lint
 lint:
-	uv run mypy --pretty src tests
-	uv run ruff check src tests
-	uv run ruff format --check src tests
+	mypy --pretty src tests
+	ruff check src tests
+	ruff format --check src tests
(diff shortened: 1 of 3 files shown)


rumpelsepp:
You need to run `uv lock` and commit it, otherwise the `uv` dev setup is broken. For now, this requires `uv sync --all-extras`, to setup a complete development setup managed by `uv`.

rumpelsepp (approved):
Approved, please run `uv lock` and commit it before merging.

ferdinandjarisch:
> Approved, please run `uv lock` and commit it before merging.

Added a fixup commit; feel free to apply it accordingly and merge it at your convenience!

Many thanks!

PR #647, opened 2024-12-11
Fix: Do not crash with stack trace on expectable error

ferdinandjarisch:


src/gallia/utils.py
```diff
@@ -351,7 +351,11 @@ def net_if_addrs() -> list[Interface]:
     if sys.platform != "linux":
         raise NotImplementedError("net_if_addrs() is only supported on Linux platforms")
 
-    p = subprocess.run(["ip", "-j", "address", "show"], capture_output=True, check=True)
+    try:
+        p = subprocess.run(["ip", "-j", "address", "show"], capture_output=True, check=True)
+    except FileNotFoundError as e:
```
rumpelsepp, line 356: This exception type means, that the `ip` tool is not available, right? I would state this in the error message as such.

rumpelsepp: Ah okay, never mind. This should be included in `{e}`.

PR #651, opened 2024-12-17
Fix logging for multiple scanners in a script

### LINKED ISSUE #641: Some scanner related actions are triggered at exit of the whole program, not fitting the Gallia 2.0 Approach of calling sub scanners
Some scanner related actions are triggered at exit of the whole program, not fitting the Gallia 2.0 Approach of calling sub scanners.
An example would be the [closing of log handlers](https://github.com/Fraunhofer-AISEC/gallia/blob/master/src/gallia/log.py#L326) which currently results in having log files with multiple scanner logs if they are called from inside a script.

rumpelsepp:
``` python
import gallia
import gallia.log
import gallia.command

logger = gallia.log.get_logger("gallia")

class Scanner1(gallia.command.Script):
    def main(self) -> None:
        logger.info(f"hi {self.__class__.__name__}")
        logger.error("error")
        logger.warning("warning")
        logger.notice("notice")
        logger.info("info")
        logger.debug("debug")
        logger.trace("trace")

class Scanner2(gallia.command.Script):
    def main(self) -> None:
        logger.info(f"hi {self.__class__.__name__}")


if __name__ == "__main__":
    # Each scanner sets up its own logging setup
    # with a logger called `gallia`. Logging setup
    # via the config class, in this case all default.
    Scanner1().entry_point()
    Scanner2().entry_point()

    # Alternatively, a context manager can be used
    # for more fine grained control.
    with gallia.log.setup_logging(
        logger_name="gallia",
        stderr_level=gallia.log.Loglevel.DEBUG,
        logfile="test.log.zst",
    ) as h:
        Scanner1(logging_handler=h).entry_point()
        Scanner2(logging_handler=h).entry_point()
```

### DIFF
--- a/src/gallia/cli/gallia.py
+++ b/src/gallia/cli/gallia.py
@@ -25,9 +25,8 @@ from gallia.log import Loglevel, setup_logging
 from gallia.plugins.plugin import CommandTree, load_commands, load_plugins
 from gallia.pydantic_argparse import ArgumentParser
 from gallia.pydantic_argparse import BaseCommand as PydanticBaseCommand
-from gallia.utils import get_log_level
 
-setup_logging(Loglevel.DEBUG)
+setup_logging("gallia", Loglevel.DEBUG)
 
 
 defaults = dict[type, dict[str, Any]]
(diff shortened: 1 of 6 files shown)


rumpelsepp:
fixed since: https://github.com/Fraunhofer-AISEC/gallia/commit/a87baca6df2242065aea5f3df99d26d9ac5bcb73

PR #655, opened 2024-12-18
chore: Enable TypeAlias lint

rumpelsepp:


rumpelsepp:
Let's enable this in october when python 3.15 is out. Let's drop 3.11 at that time.

rumpelsepp:
See #882.

PR #664, opened 2025-01-14
fix(log): Always call self.close() on __exit__

ferdinandjarisch:
Tbh I do not see a reason why the if clause is here in the first place?


rumpelsepp:
This is for exception handling within the `with` block. See here: https://docs.python.org/3/reference/datamodel.html#object.__exit__

The motivation here is that if the file is somehow broken (= an exception occured), then `.close()` would also fail, thus the call is suppressed. However, the logic is reversed for some reason anyway.

Let's merge this. If some error handling is then broken, we can fix this as needed.

PR #667, opened 2025-01-17
fix: Add RND320 interface support

rumpelsepp:
The interface is also simplified by using a dict now, instead of a list.


rumpelsepp:
In #613 it was forgotten to add it to the registry table.

PR #672, opened 2025-01-23
fix: Serialize CAN frames when DLC is not set

rumpelsepp:
- **fix: Serialize CAN frame even if dlc is not set**
- **tests: Add tests for serializing CAN frames**


rumpelsepp:
See #681.

PR #677, opened 2025-02-03
fix(db): Create single task that executes DiagnosticMessage queries

ferdinandjarisch:
This commit introduces an executor task that handles queries coming from a queue. Compared to the old approach that easily generated >100k tasks per scan, this approach is way more memory friendly.

What adds to this is that previously the tasks were generated in a different event loop than the one they were awaited in, leading to dangling objects that persisted until the python interpreter closed. This commit removes these memory leaks which becomes extremely important when scanners are used "as lib".

### DIFF
--- a/src/gallia/command/base.py
+++ b/src/gallia/command/base.py
@@ -236,6 +236,8 @@ class BaseCommand(FlockMixin, ABC):
             logger.info(p.stderr.strip(), extra={"tags": [hook_id, "stderr"]})
 
     async def _db_insert_run_meta(self) -> None:
+        # TODO: This function has to call `connect` and `disconnect` in order to be safely run in an own event loop
+
         if self.config.db is not None:
             self.db_handler = DBHandler(self.config.db)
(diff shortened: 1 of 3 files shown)


rumpelsepp:
:+1: 

Two comments:

1)

IMO it would be better to give this:

```
            asyncio.Queue[
                tuple[
                    str,
                    tuple[
                        int,
                        str,
                        str,
                        float,
                        str | None,
                        str,
                        str | None,
                        float | None,
                        str | None,
                        str | None,
                        str | None,
                        str,
                    ],
                ]
            ]
```

a concrete type, e.g., a dedicated dataclass. It would be more readable I guess.


2)

Have you a testcase or similar for this? I would like to try your design with the builtin `sqlite` module and a plain `await asyncio.to_thread(…)` call. The `to_thread` can now even cancel stuff and I question the need of `aiosqlite` in gallia's codebase.

rumpelsepp (approved):

PR #685, opened 2025-02-24
fix: consider edge case of empty or space only string when parsing ra…

fkglr:
…nges

### DIFF
--- a/src/gallia/utils.py
+++ b/src/gallia/utils.py
@@ -85,6 +85,9 @@ def unravel(listing: str) -> list[int]:
     range_delimiter = "-"
     result = set()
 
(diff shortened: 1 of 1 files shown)


rumpelsepp (approved):

src/gallia/utils.py
```diff
@@ -86,6 +86,9 @@ def unravel(listing: str) -> list[int]:
     range_delimiter = "-"
     result = set()
 
+    if len(listing) == 0 or listing.isspace():
```
rumpelsepp:
```suggestion
    if listing == "" or listing.isspace():
```

PR #699, opened 2025-03-31
!feat: Remove HAS_ARTIFACTS_DIR in favor of artifacts_base config

ferdinandjarisch:
Currently, the property `HAS_ARTIFACTS_DIR` of `BaseCommand` determines whether a folder is created and artifacts are saved.

This commit removes this property in favor of the config property `artifacts_base`, such that artifacts are automatically saved, once the property is filled in the config object.

These adaptions bring two major benefits:

1. If gallia is run quickly from the CLI, it does not clutter temporary storage or require dependencies such as dumpcap.
2. When calling individual scanners from within python with gallia as a lib, the interface to decide whether and where to artifacts are saved is more straightforward.

### DIFF
--- a/src/gallia/command/base.py
+++ b/src/gallia/command/base.py
@@ -16,7 +16,6 @@ from datetime import UTC, datetime
 from enum import Enum, unique
 from pathlib import Path
 from subprocess import CalledProcessError, run
-from tempfile import gettempdir
 from typing import Any, Protocol, Self, cast
 
 from pydantic import ConfigDict, field_serializer, model_validator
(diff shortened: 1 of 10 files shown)


rumpelsepp:
I think this makes sense.

The tests need to be fixed and somehow reordered/regrouped.

ferdinandjarisch:
> The tests need to be fixed and somehow reordered/regrouped.

I added a commit that extends the test's config file with `artifacts_base` such that behavior for the tests is the same as before.

PR #706, opened 2025-04-02
Improve service scanner

ferdinandjarisch:


rumpelsepp:
Not sure if we should drop the optional power cycle thing here. There is a bug somewhere in leave session, I assume.

ferdinandjarisch:
> There is a bug somewhere in leave session

If you refer to what we have discussed: I could identify the bug and fixed it in #704 .
The bug is triggered by calling `reconnect`, which is done during `leave_session`.

The main reason I'm dropping `leave_session` here is because its behavior is hard to predict (think of plugins).
Essentially, it currently enforces ECUReset as well as DiagnosticSessionControl, which will produce errors in case there is something wrong with those two services.

My understanding is that the service scan is run rather early into a "campaign", thus I strongly advocate to make this scanner as "raw" as possible, relying on as little available services as possible.
If you want power cycles to still be present, my suggestion is to add this explicitly and similarly to `--reset` lock it behind a CLI flag.

rumpelsepp:
Ok, cool. Thanks!
Let's add a flag in an additional pull request. I open a ticket and merge this one.

Issue #689, opened 2025-02-28
ctrl+c exception


rumpelsepp:
Catched during running the bats test during developing #688 and randomly hitting ctrl+c; this seems to be a bug that is not related to the changes in #688.

```
./tests/bats/run_bats.sh
001-invocation.bats
 ✓ invoke gallia without parameters
 ✓ invoke gallia without config
 ✓ invoke gallia with config
 ✓ invoke gallia -h
 ✓ invoke hr -h
 ✓ invoke netzteil -h
 ✓ invoke cursed-hr -h
002-scans.bats
 ✓ scan services
 ✓ scan services with database
1|gallia.commands.scan.uds.services.ServicesScanner|{"init_kwargs":null,"verbose":0,"volatile_info":true,"trace_log":false,"pre_hook":null,"post_hook":null,"hooks":true,"lock_file":null,"db":"test_scan_services_with_database.sqlite","artifacts_dir":null,"artifacts_base":"/tmp/gallia","dumpcap":false,"target":"unix-lines:///tmp/vecu.sock","power_supply":null,"power_cycle":false,"power_cycle_sleep":5.0,"ecu_reset":1,"oem":"default","timeout":2.0,"max_retries":3,"ping":true,"tester_present_interval":0.5,"tester_present":true,"properties":true,"compare_properties":true,"sessions":[1,2],"check_session":true,"scan_response_ids":false,"auto_reset":false,"skip":{}}|1740754456.24712|CET||||/tmp/gallia/services_scanner/run-20250228-155416.242152|0
 ✓ scan sessions
 ✓ scan sessions thorough
 ✓ scan identifiers sid 0x22
 ✓ scan identifiers sid 0x2e
 ✓ scan identifiers sid 0x31
 ✗ scan reset
   (in test file tests/bats/002-scans.bats, line 1)
     `#!/usr/bin/env bats' failed with status 130
   Feb 28 15:54:27.402 gallia.command.uds: ECUReset failed: subFunctionNotSupported
   Feb 28 15:54:27.402 gallia.command.uds: Switching to default session
   Feb 28 15:54:27.404 gallia.command.uds: ECUReset in session 0x01 failed: subFunctionNotSupported
Feb 28 15:54:27.913 gallia.commands.scan.uds.reset [result]: 0x03: reset level found!
Feb 28 15:54:28.420 gallia.commands.scan.uds.reset: Could not reboot ECU after testing reset level 0x03
[...]

PR #733, opened 2025-07-16
fix(cli): Properly catch KeyboardInterrupt

ferdinandjarisch:
Avoid stacktraces on keyboard interrupts


ferdinandjarisch:
Closes #689

PR #735, opened 2025-07-31
fix(transport): Remove double transport reference for UDSScanner class

ferdinandjarisch:
When initializing a UDSScanner class, it sets up its own `ECU` object that re-uses `self.transport` and offers extended functionality.

It is bad practice to maintain two references two a `BaseTransport` object at two different places, thus make `self.transport` a link to `self.ecu.transport` as soon as the latter is initialized.

### DIFF
--- a/src/gallia/command/base.py
+++ b/src/gallia/command/base.py
@@ -473,8 +473,20 @@ class Scanner(AsyncScript, ABC):
         super().__init__(config)
         self.config: ScannerConfig = config
         self.power_supply: PowerSupply | None = None
-        self.transport: BaseTransport
+        self._transport: BaseTransport | None = None
         self.dumpcap: Dumpcap | None = None
+
+    @property
+    def transport(self) -> BaseTransport:
+        if self._transport is None:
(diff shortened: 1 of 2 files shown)


src/gallia/command/base.py
```diff
@@ -473,9 +473,21 @@ def __init__(self, config: ScannerConfig):
...
 
+    @property
+    def transport(self) -> BaseTransport:
+        if self._transport is None:
+            raise RuntimeError("Transport accessed before first initialization!")
+        return self._transport
+
+    @transport.setter
+    def transport(self, transport: BaseTransport) -> None:
+        if not isinstance(transport, BaseTransport):
```
rumpelsepp, line 487: Isn't this covered by mypy already?

ferdinandjarisch: Statically yes, but I wasn't sure how exactly transports are loaded as plugins and whether types are sufficiently checked there. My idea was to have another safety-net here.

Do you want it removed?

rumpelsepp: fair enough.

PR #748, opened 2025-08-29
Transport Separation: strictly separate connect() calls and inits

ferdinandjarisch:
The transport object itself is supposed to last as long as a scanner
object, which for example allows to externally create it and pass it to
the scanner object before calling `entry_point`.

The connection itself, however, can be disconnected and reconnected
multiple times, but without creating new transport objects.


ferdinandjarisch:
@rumpelsepp Ready for review and merging!

PR #750, opened 2025-08-29
chore(dumpcap): Attach dumpcap to BaseTransport

ferdinandjarisch:
This commit links the closely connected capture mechanism with Dumpcap directly to the `BaseTransport` class, where it thematically belongs.

This allows anybody who uses `Basetransport` or its childs to easily start/stop capturing. An example for this is the DoIP discoverer, which currently has no dumpcap features because of it being an `AsyncScript`.

### DIFF
--- a/src/gallia/command/base.py
+++ b/src/gallia/command/base.py
@@ -475,14 +475,14 @@ class Scanner(AsyncScript, ABC):
 
     @property
     def transport(self) -> BaseTransport:
-        if self._transport is None:
-            raise RuntimeError("Transport accessed before first initialization!")
+        assert self._transport is not None, "Transport accessed before first initialization!"
         return self._transport
 
     @transport.setter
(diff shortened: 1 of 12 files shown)


ferdinandjarisch:
Requires some more testing, but generally ready!

Pinging once testing is finished!

ferdinandjarisch:
Tested it and it appears to work quite fine.

I realized though, that while this change permits to easily start Dumpcap on a `BaseTransport`, it is not super helpful for the DoIP discoverer, because there the `DoIPTransport` is not utilized (but directly `DoIPConnection`).
Since DoIP discoverer also sends UDP traffic, however, I would say it needs special treatment anyhow...

In the great scheme of things it's still an improvement, because Dumpcap can be started/stopped more easily outside of a `Scanner`'s `setup()` method.

PR #751, opened 2025-08-29
feat(TesterPresent): Let TesterPresent only fire for inactivity

### LINKED ISSUE #30: Implement a proper tester-present background task
The current tester-present background task is broken and it lives at the wrong place, see #21.

* It uses the suppress response flag: Several ECUs ignore this flag and respond with a message. -> boom, the UDS stack is out of sync.
* The background tasks uses the transport directly and bypasses some parts of the stack.
* There are some hacks according to the mutex, e.g. when the ECU responds with BusyResponsePending.

ferdinandjarisch:
Instead of sending regularily, let TesterPresent only become active if there was no other UDS activity for a certain amount of time.

This feature can be enabled/disabled in config file and by default is disabled to retain backwards compatibility.

### DIFF
--- a/src/gallia/services/uds/ecu.py
+++ b/src/gallia/services/uds/ecu.py
@@ -338,54 +338,72 @@ class ECU(UDSClient):
 
     async def _wait_for_ecu_endless_loop(self, sleep_time: float) -> None:
         """Internal method with endless loop in case of no answer from ECU"""
-        config = UDSRequestConfig(timeout=sleep_time, max_retry=0, skip_hooks=True)
         i = -1
         while True:
             i = (i + 1) % 4
             logger.info(f"Waiting for ECU{'.' * i}")
-            try:
-                await self.tester_present(suppress_response=False, config=config)
+            if await self._send_tester_present(timeout=sleep_time) is True:
                 break
-            # When the ECU is not ready, we expect an UDSException, e.g. MissingResponse.
-            # On ConnectionError, we additionally reconnect the transport to ensure connectivity.
-            # Since Gallia converts a ConnectionError to a MissingResponse in `request_unsafe`, however,
-            # there is a need to reconnect the transport also in case of high-level UDSExceptions
-            # such as MissingResponses that are raised (__cause__) from ConnectionErrors.
-            except (ConnectionError, UDSException) as e:
-                logger.debug(f"ECU not ready: {e!r}")
-                if isinstance(e, ConnectionError) or isinstance(e.__cause__, ConnectionError):
-                    logger.debug("Reconnecting…")
-                    await self.reconnect()
         logger.info("ECU ready")
 
     async def wait_for_ecu(
         self,
-        timeout: float | None = 10,
+        timeout: float = 10,
     ) -> bool:
-        """Wait for ecu to be alive again (e.g. after reset).
-        Sends a ping every 0.5s and waits at most timeout.
-        If timeout is None, wait endlessly"""
+        """
+        Wait for ECU to be alive again (e.g. after reset) for at most `timeout`.
+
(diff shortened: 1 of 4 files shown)


src/gallia/services/uds/ecu.py
```diff
@@ -335,54 +335,73 @@ async def transmit_data(
...
-            logger.info("connection lost; tester present waiting…")
+            response = await self.tester_present(
+                config=UDSRequestConfig(
+                    max_retry=0,
+                    tags=["tp"],
+                    timeout=timeout,
+                    skip_hooks=True,
+                )
+            )
+            raise_for_error(response)
```
ferdinandjarisch: This results in endless waits if the response to TesterPresent is not positive. Needs to be looked into before merging!

ferdinandjarisch: Error is no longer raised!

src/gallia/services/uds/ecu.py
```diff
@@ -525,3 +527,90 @@ async def _request(
...
+        """Call this whenever you send or receive real traffic to reset the timer"""
+        if self._ignore_activity is True:
+            return
+        self._activity_event.set()
+
+    async def start(self) -> None:
+        self._cancelled = False
+
+        if self._task is not None:
+            logger.warning("TesterPresentSender already running")
```
rumpelsepp, line 568: If `.start()` is called multiple times, isn't this a programming error? Maybe we could raise an `AssertionError` instead of logging a warning.

src/gallia/services/uds/ecu.py
```diff
@@ -349,92 +338,100 @@ async def transmit_data(
...
+        await self.tester_present_task.start()
+
+    async def detach_tester_present_sender(self) -> None:
+        logger.debug("Detaching TesterPresentSender")
 
-    async def stop_cyclic_tester_present(self) -> None:
-        logger.debug("Stopping tester present worker")
         if self.tester_present_task is None:
-            logger.warning("BUG: stop_cyclic_tester_present() called but no task running")
+            logger.warning("Attempted to detach TesterPresentSender, but it is already gone!")
```
rumpelsepp, line 430: Same here, isn't this a programming error? Maybe raise AssertionError (cf. below)?

rumpelsepp:
@ferdinandjarisch Not sure about the assertion stuff I suggested. In Python, there is `__debug__` which is `True` when Python is not started with `-O`. When `__debug__` is `False` (in other words: python is started with `-O`), then all `assert` statements are optimized out. This could mean, that we would see such assertion errors only in testing mode (e.g., in the CI or during development). 

However, this seems to be rather unused… I tested a few Python tools on my Debian and there `__debug__` is always true. Not sure if a warning is better here or we accept raising `AssertionErrors` in this case.

Docs according to `__debug__`: https://docs.python.org/3/library/constants.html#debug__

----

Rest looks good; let's briefly discuss how we handle assertions and then let's merge.

ferdinandjarisch:
Apart from your point about asserts in python, I see them as hard checks to abort execution if certain, mandatory conditions are not met.
This makes sense, for example, when checking whether a transport exists during the set-up of a `UDSScanner` object.
In this case, however, starting and stopping the `TesterPresentSender` is something very dynamic, that within gallia might only be done at a few places, but within plugins could be done multiple times.

Thus, I prefer `start` and `stop` to be "forgiving" calls, that if already started or stopped before, essentially become no-ops apart from a logging message, instead of crashing the entire application with an `AssertionError`.

Attaching or detaching `TesterPresentSender` arguably leans more towards assertion, but again I would advocate for making our code "forgiving", keeping in mind that our main purpose is a "penetration testing framework", where flexibility is key!

rumpelsepp:
Fair enough!

PR #762, opened 2025-09-11
Add more base classes for more fine grained parameters

fkglr:
This adds two new base classes between the existing AsyncScript and Scanner hierarchy.

Basically what happens is:
- The `lock_file` parameter from `GalliaBaseModel` gets moved down to the new `LockableScript`, to avoid unnecessary locks in any scripts if the user uses a config file with a lock file set
- The `power_supply`, `power_cycle` and `power_cycle_sleep` parameters get moved up to the new `MetaScanner` class, to also allow discovery scripts to make use of the powercycle functionality automatically and make further use of the power supply if necessary
- Some scripts have been adjusted to use one of the new base classes as their parent class

### DIFF
--- a/src/gallia/cli/netzteil.py
+++ b/src/gallia/cli/netzteil.py
@@ -7,8 +7,7 @@ from typing import Any, Literal, Self
 from pydantic import field_serializer, model_validator
 
 from gallia.cli.gallia import parse_and_run
-from gallia.command import AsyncScript
-from gallia.command.base import AsyncScriptConfig, ScannerConfig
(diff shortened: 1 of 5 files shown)


ferdinandjarisch (commented):
Instead of implementing further classes for more fine-grained parameters, I advocate for the following:

1. Remove the Locking-Feature alltogether: #776 
2. Merge classes for easier maintenance and less complexity: #774 

When following this proposal, this MR should be closed.

rumpelsepp:
This is addressed differently by removing flock alltogether, see #776

PR #769, opened 2025-10-08
fix(deps): Limit max pydantic version for compatibility

ferdinandjarisch:


rumpelsepp:
For now acceptable, however this breaks support for Python 3.14 and need to be addressed.

see also #745 
/cc @fkglr

PR #770, opened 2025-10-08
fix(xcp): Add missing type hints

ferdinandjarisch:


rumpelsepp:
This file is autogenerated from upstream. Upstream updated the type definitions a few times in the past. There are two solutions: a) File a merge request upstream or b) let's just drop a # type: ignore in the whole file. I vote for the latter.

rumpelsepp:
Solution b) is already addressed here: https://github.com/Fraunhofer-AISEC/gallia/blob/48afcdca6031b44688e2c3d74f4ea6d54174dbb2/pyproject.toml#L82-L84

Is this setting somehow ignored on your machine?

ferdinandjarisch:
> Solution b) is already addressed here:
> 
> https://github.com/Fraunhofer-AISEC/gallia/blob/48afcdca6031b44688e2c3d74f4ea6d54174dbb2/pyproject.toml#L82-L84
> 
> Is this setting somehow ignored on your machine?

Since mypy does not support merging of config files this newly added ignore is indeed not taken into account...

rumpelsepp:
Ah, I see.

PR #771, opened 2025-10-09
Fix(pydantic): Store original field info to work around changes in pydantic 2.12

fkglr:
pydantic 2.12 introduces a change that discourages using subclasses of the FieldInfo class by making it final and often replaces the original object with a new one of the FieldInfo class. As a consequence, the extra information of any object of a subclass is lost.

This commit therefore stores the original FieldInfos during class instantiation in the _original_fields variable. These fields can then be used by the parser to retrieve any additional information that is not present in the potentially changed fields.

This commit does not yet address the inheritance problem.


fkglr:
Added to #745

PR #772, opened 2025-10-10
fix(doip/hsfz): Correctly handle missing values in `IntEnum`s

ferdinandjarisch:


rumpelsepp:
There is code duplication that might be overlooked in future updates/fixes/…. Maybe use a MixIn and reuse your snippet over all enums?

rumpelsepp:
Thx. Is this ready to merge from your side?

ferdinandjarisch:
> Thx. Is this ready to merge from your side?

Not yet! Some minor changes and a final squash are still due!
Further, I wanted to ask what you think about this, but I'll interpret your question as "go ahead"! :smiley:

rumpelsepp:
When this works, I think the code is fine. I would suggest to make your `IntEnumWithMissing` private by renaming it to `_IntEnumWithMissing`. Otherwise looks legit!

ferdinandjarisch:
Ready to merge!

PR #776, opened 2025-10-16
!chore: Remove flock support

rumpelsepp:
Our builtin flock support is broken when gallia is used as a library.
However, this is a dropin replacement at the shell level:

        $ alias gallia="flock /tmp/gallia.lock gallia"


ferdinandjarisch:
I fully support this removal, LGTM!

Specifically @fkglr for awareness!

PR #803, opened 2025-11-19
tests: Add infrastructure to unit test scanner modules

rumpelsepp:
This one creates the required infrastructure to test scanning modules via pytest. The QueueTransport is needed as an in memory networking layer. Our UDSServer module is used to implement test specific behaviour of a DuT.

old:

Found this locally and almost deleted it during chore. This superseeds #496, but is very much WIP and not ready.


rumpelsepp:
I would like to bring this in for 2.0. We can extend the test cases easily after that.
@ferdinandjarisch Any objections? If not, I would merge this and try to get gallia 2.0 released this month.

PR #783, opened 2025-10-17
chore: rm -rf xcp

rumpelsepp:

PR #807, opened 2025-11-27
feat: Improve ISOTP-Discovery

ferdinandjarisch:


ferdinandjarisch:
Ready for review, but needs final testing on a real world target.

rumpelsepp:
I fear that in a future refactor/change one of these checks will be missed: `msg.arbitration_id > CAN_SFF_MASK or self.config.force_extended_ids`.

What about putting this into a helper and using that instead?

```python

def _should_use_extended(self, arbitration_id: int) -> bool:
    return arbitration_id > CAN_SFF_MASK or self.force_extended_id
```

Then you can replace this code

```python
if msg.arbitration_id > CAN_SFF_MASK or self.config.force_extended_ids:
    # foo
```

with something like this:

``` python
if self._should_use_extended(msg.arbitration_id):
    # foo
```

src/gallia/transports/can.py
```diff
@@ -100,21 +96,20 @@ def unpack(cls, frame: bytes) -> Self:
         #   #define CAN_EFF_FLAG 0x80000000U /* EFF/SFF is set in the MSB */
         #   #define CAN_RTR_FLAG 0x40000000U /* remote transmission request */
         #   #define CAN_ERR_FLAG 0x20000000U /* error frame */
-        is_extended_frame_format = bool(can_id & CAN_EFF_FLAG)
+        is_extended_frame_format = bool(can_id & s.CAN_EFF_FLAG)
```
rumpelsepp: Please stick to using our constants. We had several setups already where a Python interpreter without these constants was used (happens occasionally on embedded Linux). These constants are a `./configure` option during building python. By just shipping our own constants (which are autogenerated) we keep these platforms working for free AND fix differences between python versions, too.

ferdinandjarisch: I adapted those constants to [ISOTP](https://github.com/Fraunhofer-AISEC/gallia/blob/master/src/gallia/transports/isotp.py#L100-L103), which is the actual can-related transport used throughout gallia.

Obviously it was no problem up until now, so I cannot understand how it might become a problem for can-raw?

src/gallia/transports/isotp.py
```diff
@@ -31,9 +31,9 @@
 
 
 class ISOTPConfig(BaseModel):
-    src_addr: int
```
rumpelsepp, line 34: The Linux kernel uses tx/rx terminology, see here: https://www.kernel.org/doc/html/next//networking/iso15765-2.html#basic-node-example

The `isotpsend` tooling uses "source", "destination" terminology, see here: https://github.com/linux-can/can-utils/blob/88a0417a6b4ad2d16231f08ee659218c81540bd1/isotpsend.c#L67-L68
That one was historically picked up by me in the first implementation.

The ISO-TP standard uses "source", "target" terminology: https://netcult.ch/elmue/HUD%20ECU%20Hacker/ISO%2015765-2.pdf

I am not aware where the suggested terminology (req/res) was invented, however, I would suggest to not create an additional terminology here. If you prefer streamlining the config keys, I suggest sticking to either the kernel or the iso standard. I prefer the kernel one, since that's the interface we use in this transport.

src/gallia/transports/can.py
rumpelsepp: Then it was just missed, or not triggered by the mentioned setup. For consistency, we should use all our own constants. So adapt it the other way round.

src/gallia/transports/isotp.py
ferdinandjarisch: The current usage of `src_addr` and `dst_addr` is very confusing and even changes midway through the code of the ISOTP discoverer currently on master (c.f. [exhibit 1](https://github.com/Fraunhofer-AISEC/gallia/blob/master/src/gallia/commands/discover/uds/isotp.py#L138-L140), [exhibit 2](https://github.com/Fraunhofer-AISEC/gallia/blob/master/src/gallia/commands/discover/uds/isotp.py#L180-L181C59))...

`req_id` and `res_id` better highlights the difference in outgoing UDS-requests and incoming UDS-responses.
Calling it `tx_id` and `rx_id` instead matches that approach 100% and is fine by me.

Funnily enough, your [second source also immediately falls back to `tx_id` and `rx_id`](https://github.com/linux-can/can-utils/blob/88a0417a6b4ad2d16231f08ee659218c81540bd1/isotpsend.c#L107-L117).

rumpelsepp: > Funnily enough, your [second source also immediately falls back to tx_id and rx_id](https://github.com/linux-can/can-utils/blob/88a0417a6b4ad2d16231f08ee659218c81540bd1/isotpsend.c#L107-L117).

That's because of the kernel interface and the relevant c structures in the uapi header here: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/include/uapi/linux/can.h#n252

The suggested variants (`req_id` and `res_id`) do not contribute to a better understanding, since it still get's confusing when you know the kernel interface or the other mentioned ones.

Why not just use source and target? We have that in DoIP already and the ISO-TP standard uses this terminology. To make my point clear: I am strongly against creating a new and different terminology that only gallia uses.

ferdinandjarisch: > That's because of the kernel interface and the relevant c structures in the uapi header here: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/include/uapi/linux/can.h#n252

Perfect, then let's use `tx_id` and `rx_id`, as I already agreed before! :confetti_ball:

src/gallia/transports/can.py
ferdinandjarisch: Change was reverted!

src/gallia/transports/isotp.py
rumpelsepp: Cool, thanks!

rumpelsepp:
So far ACK, please merge once you are happy with your real world testing.

rumpelsepp (approved):

ferdinandjarisch:
> So far ACK, please merge once you are happy with your real world testing.

Thanks, will do! :smiley: 

Just for you not to miss it: I added two small commits (ebadde0b73999ec9a1e5f02215072ca55c6434d2, 6b4911fe4dfa10a0a35417cb28ea4f309e530a42). The latter together with #783 would allow to remove `can_id_repr()` altogether, since it is only used in `IsotpDiscoverer` and `FindXCP`.

rumpelsepp:
Ah, yes. Cool, that's fine. The `*_repr()` stuff already caused me some headaches in the past due to cyclic dependencies. Happy to see one of those guys is going to die :gun:.

ferdinandjarisch:
> Ah, yes. Cool, that's fine. The `*_repr()` stuff already caused me some headaches in the past due to cyclic dependencies. Happy to see one of those guys is going to die 🔫.

In that case, I went all-in and already removed them altogether: 86e513e159551f2bcd4051c53ff199bed232b3a7

PR #815, opened 2025-12-01
linters: Add experimental CI job for ty

rumpelsepp:
ty will be the next big thing: https://docs.astral.sh/ty/

Enable it, but make it entirely optional. We see test results, but it
does not block our CI if it fails; and it will fail in the beginning. I
hope that it reveals some bugs that mypy misses.


rumpelsepp:
Included in #820

PR #817, opened 2025-12-01
refactor: Remove transport property of UDSScanner

### LINKED ISSUE #808: Remove transport property of `UDSScanner`
I think the `UDSScanner` does not necessarily need to have a reference to the transport, since only `ECU` is using it.

rumpelsepp:
fixes #808


rumpelsepp:
@ferdinandjarisch Seems to be straight forward. Linters and bats tests pass.

ferdinandjarisch (changes requested):
NACK

This MR currently removes the interface to provide an external transport to Scanner objects.

PR #821, opened 2025-12-07
feat(sa_dump_seeds): Improved handling of reset and sleep

emeisd:
- Prints a note when resets slow down automatic key-length detection
- Improves sleep between consecutive seed requests in cases where the main loop is `continue`-ed early, before the current `sleep` is even reached.
- Minor corrections in scanner description

### DIFF
--- a/src/gallia/commands/scan/uds/sa_dump_seeds.py
+++ b/src/gallia/commands/scan/uds/sa_dump_seeds.py
@@ -57,10 +57,10 @@ class SASeedsDumperConfig(UDSScannerConfig):
 
 
 class SASeedsDumper(UDSScanner):
-    """This scanner tries to enable ProgrammingSession and dump seeds for 12h."""
+    """This scanner tries to dump SecurityAccess seeds."""
 
     CONFIG_TYPE = SASeedsDumperConfig
-    SHORT_HELP = "dump security access seeds"
+    SHORT_HELP = "dump SecurityAccess seeds"
 
     attempt_reset: bool = False
     is_key_length_determined: bool = False
@@ -166,6 +166,18 @@ class SASeedsDumper(UDSScanner):
             # Start with length 1 in automatic search
             self.key_length = 1
 
+            # When resets are only triggered on demand (self.config.reset == 0), their impact on the
+            # time required for key length determination is usually minimal. Thus, the following
+            # notice should only be printed if resets are explicitly forced every n-th attempt by
+            # the user.
+            if (self.config.reset is not None) and (self.config.reset > 0):
+                logger.notice(
+                    "You combined '--reset N' with '--send-zero-key' without providing a key "
+                    "length. Automatically determining the key size expected by the ECU in this "
+                    "setting might be very slow. Consider not forcing resets or set the key length "
+                    "explicitly if it is known."
+                )
+
         while duration <= 0 or time.time() - start_time < duration:
             # Print information about current dump speed every `interval` seconds.
             # As request/response times can jitter a few seconds, we 'arm' the print
@@ -174,9 +186,15 @@ class SASeedsDumper(UDSScanner):
             i = int(time.time() - start_time) % interval
             if i >= interval // 2:
                 print_speed = True
-            elif i < interval // 2 and print_speed is True:
+            # Print dump speed only when actually dumping seeds, not while determining the key size,
+            # as this will lead to confusing output.
+            elif i < interval // 2 and print_speed is True and self.is_key_length_determined:
                 self.log_size(seeds_file, time.time() - start_time)
                 print_speed = False
+
+            if self.config.sleep is not None:
+                logger.info(f"Sleeping for {self.config.sleep} seconds between seed requests…")
+                await asyncio.sleep(self.config.sleep)
 
             if self.config.reset is not None:
                 if (self.config.reset == 0 and self.attempt_reset) or (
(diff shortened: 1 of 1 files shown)


src/gallia/commands/scan/uds/sa_dump_seeds.py
```diff
@@ -202,8 +204,19 @@ async def main(self) -> None:
                     logger.error(f"ECU persistently lost session {g_repr(self.config.session)}")
                     sys.exit(1)
 
+            if (
+                (self.config.sleep is not None)
+                and (last_seed_req is not None)
+                and ((delta := self.config.sleep - (time.time() - last_seed_req)) > 0)
```
rumpelsepp: This is quite hard to understand. Maybe changing the type of `last_seed_req` to `float` and initializing to `0` instead makes it more readable.

Then your code looks like this:

```python
last_seed_req = 0

# …

# The first calculation is negative.
delta = self.config.sleep - (time.time() - last_seed_req)

if self.config.sleep is not None and delta > 0:
    # …
```

ferdinandjarisch: I do not see the benefit of this complex sleep functionality anyhow. Why should we keep track of a delta?

ferdinandjarisch:
Further, it was discussed today that it might make sense to have the `--reset` flag be respected, also during the automatic determination of the seed length, e.g. if the ECU already locks out during that part of the script.

@emeisd What do you think about this? Would it be sufficient to maybe give a hint with notice level to provide a key length in order to skip a potentially slow auto-determination?

src/gallia/commands/scan/uds/sa_dump_seeds.py
emeisd: This is now heavily simplified. The delta is not calculated anymore.

emeisd:
> Further, it was discussed today that it might make sense to have the `--reset` flag be respected, also during the automatic determination of the seed length, e.g. if the ECU already locks out during that part of the script.
> 
> @emeisd What do you think about this? Would it be sufficient to maybe give a hint with notice level to provide a key length in order to skip a potentially slow auto-determination?

Resets during key length determination are now allowed again and instead the suggested note is printed.

src/gallia/commands/scan/uds/sa_dump_seeds.py
```diff
@@ -166,14 +166,24 @@ async def main(self) -> None:
...
             # Print information about current dump speed every `interval` seconds.
             # As request/response times can jitter a few seconds, we 'arm' the print
             # in one half and 'shoot' once in the other half.
             interval = 60
             i = int(time.time() - start_time) % interval
             if i >= interval // 2:
-                print_speed = True
+                # Print dump speed only when actually dumping seeds, not while determining the
+                # key size, as this will lead to confusing output.
+                print_speed = self.is_key_length_determined
```
ferdinandjarisch: This feels wrong!

What you are trying to achieve is to add another conditional check to the line below, but you are hiding it in the value of `print_speed`.

I would suggest to leave it like this, and instead add the comment to the conditional check and change it to:

```python
                elif self.is_key_length_determined and print_speed is True and i < interval // 2:
```

src/gallia/commands/scan/uds/sa_dump_seeds.py
```diff
@@ -202,6 +212,10 @@ async def main(self) -> None:
                     logger.error(f"ECU persistently lost session {g_repr(self.config.session)}")
                     sys.exit(1)
 
+            if self.config.sleep is not None:
```
ferdinandjarisch: I would move it even further up, basically directly after the printing stuff and before the reset.

Like this, it might have unintended side effects, if sleeping for too long after resetting/changing the session.

src/gallia/commands/scan/uds/sa_dump_seeds.py
```diff
@@ -166,14 +166,24 @@ async def main(self) -> None:
             # Start with length 1 in automatic search
             self.key_length = 1
 
+            if self.config.reset is not None:
```
ferdinandjarisch: I have not tested this, but simply providing `--reset` without a value does not trigger reconnects in the automatic key length determination, I believe?

If that is the case, change it to an explicit truthy check to only print this message for "forced" resets.


```suggestion
            if self.config.reset:  # Explicit truthy check for !(None || 0)
```

ferdinandjarisch (changes requested):

src/gallia/commands/scan/uds/sa_dump_seeds.py
emeisd: Changed it as discussed internally. Your primary concerns regarding session losses are addressed and also the further discussed concerns of loosing possibly desired functionality if sleeping before resets.

src/gallia/commands/scan/uds/sa_dump_seeds.py
emeisd: You were right. Adjusted it so that it now is functionally equivalent to your suggestion, but the conditions and the considerations behind them are easier to understand at first glance.

rumpelsepp:
Linter fails were temporary, caused by sth. on Github's side. I re-ran them, fixed now.

src/gallia/commands/scan/uds/sa_dump_seeds.py
emeisd: Changed it again after further discussion. It functions now purely as a loop delay.

ferdinandjarisch: LGTM, thanks!

ferdinandjarisch (approved):
Looks good to me, thanks!

Commits should be squashed latest when merging!

PR #828, opened 2025-12-17
Fix handling of NRC 0x13 during service scan

ferdinandjarisch:


src/gallia/commands/scan/uds/services.py
```diff
@@ -192,13 +199,24 @@ async def perform_scan(self, session: None | int = None) -> tuple[dict[int, Any]
...
                 logger.result(f"{g_repr(sid)}: available in session {g_repr(session)}: {resp}")
                 result[sid] = resp
                 break
 
+            else:
+                if at_least_one_valid_response is not None:
+                    logger.result(
+                        f"{g_repr(sid)}: probably available in session {g_repr(session)}: {at_least_one_valid_response}"
+                    )
+                    result[sid] = at_least_one_valid_response
```
rumpelsepp, line 225: Should `result` store the "probably available" and the, let's call it, "confirmed available" services separately?

fkglr: From my point of view, a service, which returns incorrectMessageLengthOrInvalidFormat should in general be seen as available. This loop over different payload length tries to reduce false positives in potentially rare cases, where the behavior is different than depicted in the standard, but I don't see any reason to complicate the behavior any further.

ferdinandjarisch: I do not see a benefit in this, in the way we are currently working with the result.

fkglr (approved):
Tested using the vecu

ferdinandjarisch:
Tested on physical ECU as well...

PR #829, opened 2025-12-19
Improve CAN-FD handling and remove unused XCP-related code

ferdinandjarisch:


ferdinandjarisch:
@rumpelsepp what is the issue with the linter?

rumpelsepp:
```
* Unused licenses: LGPL-3.0-only

…

* Fix unused licenses: At least one of the license text files in 'LICENSES' is
  not referenced by any file, e.g. by an 'SPDX-License-Identifier' tag. Please
  make sure that you either tag the accordingly licensed files properly, or
  delete the unused license text if you are sure that no file or code snippet is
  licensed as such.
```

Delete the LGPL license in the LICENCES directory.

rumpelsepp:
Also, when you delete the xcp stuff, then the dependency `construct` and `construct-types` (oder so ähnlich) is not required any more and can be dropped.

rumpelsepp:
I took the liberty to fix this quickly.

src/gallia/transports/can.py
```diff
@@ -149,8 +149,8 @@ async def connect(
         sock = s.socket(s.PF_CAN, s.SOCK_RAW, CAN_RAW)
         sock.bind((self.target.hostname,))
 
-        if self.config.is_fd is True:
-            sock.setsockopt(SOL_CAN_RAW, CAN_RAW_FD_FRAMES, 1)
+        # Enable reception of CAN FD frames in addition to classical frames
+        sock.setsockopt(SOL_CAN_RAW, CAN_RAW_FD_FRAMES, 1)
```
rumpelsepp: This makes CAN devices that support CAN_FD a hard requirement for using gallia. To my knowledge, this call fails when the attached CAN device does not support CAN FD. Consequently, this locks out, e.g., raspberry pi based setups that use a CAN shield which (as to my knowledge) just support legacy CAN.

I am against such a breaking change, since I know at least some setups existed. Can we perhaps just catch `OSError` here and log a warning?

ferdinandjarisch: Exactly my thought: Wrap it in a try block, catch the proper Error, and emit a warning that FD is not supported.

Is `OSError` the correct one?

ferdinandjarisch:
I just compared against #783 : the important parts are included, meaning this MR fully supersedes it

src/gallia/transports/can.py
rumpelsepp: IIRC yes. There is the `.errno` attribute for the actual error code and `.strerror` containing the error string. Perhaps this is useful: https://docs.python.org/3/library/exceptions.html#OSError.strerror

ferdinandjarisch: Try block is implemented.

ferdinandjarisch:
I pushed a proper rebase. Should be tested one more time by @fkglr though.

rumpelsepp (approved):

ferdinandjarisch:
Tested and ready to merge.

PR #837, opened 2026-01-15
chore: Try to make dependabot less verbose

rumpelsepp:
This should instruct dependabot to create only one merge request for all dependencies (per ecosystem, but github actions are not so frequently updated). Then we can easily hit the squash button prior to merge.


ferdinandjarisch:
Let's hope it works, thanks! :+1:

PR #838, opened 2026-01-15
chore: Remove platformdirs dependency

rumpelsepp:
This dependency is only used on two locations in the gallia code. Let's
remove it and do the trivial tasks ourselves. Let's follow Russ Cox'
advice here. Also reduces dependabot noise.

Russ Cox says:

> A little copying is better than a little dependency

https://research.swtch.com/deps


ferdinandjarisch:
Hm, can this be avoided altogether, e.g. by just dropping the two references?

Tempdir should work fine without a `dir` (possibly not profiting from caches?), and the test itself I feel does not justify having it...

ferdinandjarisch:
> possibly not profiting from caches?

I did not properly look at the code in the first place: this is not even about caching, but about a "source directory" which is instead the users's cache directory? CONFUSING

ferdinandjarisch (changes requested):
See previous comments: I would drop it altogether.

rumpelsepp:
According to `~/.cache/gallia`: We had this in the past. Without `dir=` logfiles are unzipped to `/tmp` which is often a tmpfs in RAM. We moved this to the user cache on disk to avoid OOM issues.

According to `~/.config/gallia/config.toml`: I tend to prefer keeping this as well, since this is a clearly defined location where people can start using gallia. Not all people might start with a full fledged project based folder structure as we do internally.

For the latter, I forgot to commit this in the first place, it is here: https://github.com/Fraunhofer-AISEC/gallia/blob/6fd1359485df4d63115ed10145ca142710aeda62/src/gallia/config.py#L11

PR #842, opened 2026-01-28
Cli features

ferdinandjarisch:


src/gallia/command/uds.py
```diff
@@ -104,6 +105,25 @@ def check_power_supply_required(self) -> Self:
 
         return self
 
+    @model_validator(mode="after")
+    def replace_target_alias(self) -> Self:
+        if len(self.target.url.scheme) == 0 and self.db is not None:
+            asyncio.run(self._replace_target_alias())
```
rumpelsepp: Not a fan of starting a whole event loop including database access in type magic. We should really stick to using pydantic for *validation* and not *manipulation*. 

Have you looked at putting this in `.setup()` instead?

https://github.com/Fraunhofer-AISEC/gallia/blob/73877e1f33a82a45a6901ec84ed05a7df58aaa32/src/gallia/command/uds.py#L172

ferdinandjarisch: I thought about putting it into `setup` in the first place, but due to some reason I cannot remember I instead used the validator approach.

Added a fixup commit for the `setup` way.
Let's see if I can recall the reason in due time...

ferdinandjarisch: I remembered! :confetti_ball: 

The reason for my change being in the validator was that it has to be there, if `TargetURI` actually puts constraints on its `__init__` argument. At the moment, however, this is not the case, which is why the `setup` approach also works.

To give more detail:

Currently, `TargetURI` is nothing more than a string-wrapper on init:

```python
    def __init__(self, raw: str) -> None:
        self.raw = raw
        self.url = urlparse(raw)
        self.qs = parse_qs(self.url.query)
```

The actual "check", whether it is a usable URI for scans, is then hidden in the property `scheme` and is not executed until this property is accessed:

```python
    @property
    def scheme(self) -> TransportScheme:
        """The URI scheme"""
        return TransportScheme(self.url.scheme)
```

Luckily, `property` is not accessed until the transport is loaded in `load_transport`.

Overall, the current implementation feels weird and I expected that if the "check" for TransportSchemes is moved some day, funny things might happen. Thus, I moved it directly to the model validator.

Having written all of this, I wonder if the transport scheme check should be removed from `TargetURI.scheme` anyhow.
If you disagree, be invited to think about how "file:///dev/RND_<serial>?product_id=RND320" can be a valid `PowerSupplyURI`.

rumpelsepp: Ahh. Yes this is kind of weird. I think we can just remove that weird check.

ferdinandjarisch: Added a commit that removes it.

PR #845, opened 2026-02-03
fix(cursed-hr): use wcwidth to process unicode

fkglr:
Currently cursed-hr fails when parsing unicode sequences (depending on window size).
This PR fixes it by using the wcwidth library's wrap function.
At least wcwidth 0.3 is required for the wrap function.


rumpelsepp:
This is required due to an upstream bug: https://github.com/python/cpython/issues/56777

PR #849, opened 2026-02-04
fix: Fix upcoming ruff 0.15 lints

rumpelsepp:


rumpelsepp:
@ferdinandjarisch In case you need this internally, we can merge this ahead of the next dependabot action.

ferdinandjarisch (approved):
I already have these changes locally and as far as I see it, they are independent of dependabot merges.

So go ahead and merge it!

PR #850, opened 2026-02-04
fix: Migrate from urlparse() to urlsplit()

### LINKED ISSUE #841: urlparse() is obsolete
https://github.com/python/cpython/issues/144148

We should move to urlsplit(), since urlparse() is appearently based on an obsolete RFC.

rumpelsepp:
Fixes #841

### DIFF
--- a/src/gallia/commands/discover/doip.py
+++ b/src/gallia/commands/discover/doip.py
@@ -7,7 +7,7 @@ import socket
 import sys
 from collections.abc import Iterable
 from itertools import product
-from urllib.parse import parse_qs, urlparse
+from urllib.parse import parse_qs, urlsplit
 
 from gallia.command import AsyncScript
 from gallia.command.base import AsyncScriptConfig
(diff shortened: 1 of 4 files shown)


ferdinandjarisch (approved):
Is it as easy as that? No differences between `ParseResult` and `SplitResult` we need to take care of?

rumpelsepp:
Indeed it seems to be that simple. The difference is that `urlsplit()` uses the current RFC for parsing. :shrug:

rumpelsepp:
@ferdinandjarisch I merge it. If your tests start to suddenly fail, I will revert this and investigate further. From my current understanding everything should be the same, except that it is based on the updated RFC. :tm:

ferdinandjarisch:
> @ferdinandjarisch I merge it. If your tests start to suddenly fail, I will revert this and investigate further. From my current understanding everything should be the same, except that it is based on the updated RFC. ™️

`params` are no longer part of it, but afaik this is not used by us, so all good it seems!

rumpelsepp:
Nobody uses params tbh. I used params in Python2 when I was a student in the lab, IIRC it was sth. like this `schema://path?key=value;param` (the `param` part in my example). This just disappeared from the libraries over the years.

PR #853, opened 2026-02-20
Hacky way of removing the need to type Literal for AutoLiteral

fkglr:


ferdinandjarisch (approved):
Looks fine to me, but I am by far no expert in this typing business..

PR #855, opened 2026-03-06
feat(config): Add `create_template` method to `GalliaBaseModel`

ferdinandjarisch:
This method generates a string which can be loaded as config value.

When called on an "empty" `GalliaBaseModel`, this string represents a template config, and if called on an instantiated config with custom values, those values are represented in the config file.

### DIFF
--- a/src/gallia/command/config.py
+++ b/src/gallia/command/config.py
@@ -3,13 +3,14 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import binascii
-import os
+import json
 import tomllib
 import typing
 from abc import ABC
 from collections.abc import Callable
 from enum import Enum
 from pathlib import Path
+from textwrap import wrap
 from typing import (
     TYPE_CHECKING,
     Annotated,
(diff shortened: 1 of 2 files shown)


ferdinandjarisch:
Ready for review

src/gallia/command/config.py
```diff
@@ -438,7 +438,9 @@ def attributes_from_config(cls, config: Config, source: str = "config file") ->
             for name, info in original_infos.items():
                 if isinstance(info, ConfigArgFieldInfo):
                     config_attribute = (
-                        f"{info.config_section}.{name}" if info.config_section != "" else name
+                        f"{info.config_section}.{name}"
+                        if info.config_section not in ("", None)
+                        else name
```
rumpelsepp: I think it could be `f"{info.config_section}.{name}" if info.config_section else name`

rumpelsepp (approved):
Looks good. Check my comment according to the `if` thingy.

src/gallia/command/config.py
ferdinandjarisch: Yes, it could be!

However, for the plugins we agreed to avoid the truthy check whenever possible because we already had multiple bugs due to unintended side effects, mainly though with `if 0` being `False`.

What do you prefer here?

rumpelsepp: true, but following this pattern, here is another quirk:

`if foo not in ("", None)` bails down to `if foo != "" and foo != None` which is wrong. It should be `is not None`. However, in this particular case I think `if foo` is prettier and raises less questions.

ferdinandjarisch: changed accordingly

PR #858, opened 2026-03-31
Config changes

fkglr:


ferdinandjarisch:
Slightly reworked and functionally contained in #855

PR #877, opened 2026-07-20
chore: Remove SOCKS proxy support

rumpelsepp:
This is now completely untested and I assume it might be horribly
broken.

### DIFF
--- a/src/gallia/dumpcap.py
+++ b/src/gallia/dumpcap.py
@@ -5,7 +5,6 @@
 import asyncio
 import gzip
 import io
-import os
 import shutil
 import socket
 import struct
@@ -14,7 +13,6 @@ from asyncio import subprocess
 from datetime import datetime
 from pathlib import Path
 from typing import Self, cast
-from urllib.parse import urlsplit
 
 from gallia.log import get_logger
 from gallia.utils import handle_task_error, set_task_handler_ctx_variable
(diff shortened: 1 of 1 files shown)


rumpelsepp:
@ferdinandjarisch I found this while working on #876. I think we should remove the SOCKS thing. I used this in COVID days for running gallia through an SSH tunnel. Just to make sure: Does anybody still use this?

Note: Keeping the call into getaddrinfo() still makes sense, since it handles cases such as resolving hostnames or even portnames, and ip aliases and all sorts of cornercases you can't even imagine…

ferdinandjarisch:
>Does anybody still use this?

No, the tools are all run from local machines directly connected to the targets via SSH.
IMHO both from a performance-perspective as well as from a continuity-perspective this set-up has substantial drawbacks.

Since the change is only in `dumpcap.py` I wonder, however, if this is just a left-over or if there is more hidden functionality connected to SOCKS somewhere else?


>Keeping the call into getaddrinfo() still makes sense

AFAIK this is not really used either, since all ETH-based connections are plain IPv4 addresses. I am not even sure if IPv6 is supported...
So from my perspective it could be dropped, but since it is there already, we can also keep it... :shrug:

rumpelsepp:
to SOCKS: I agree, I think there is not more stuff according to SOCKS in the codebase. IIRC the only really problem was capturing the correct traffic.

to getaddrinfo(): I would leave this as is, since from my experience this one fixes more than it breaks. It is a noop for plain addresses and ports, but does its magic once you start using hostnames and so on.

PR #878, opened 2026-07-20
chore: Cleanup obsolete development deps

rumpelsepp:
Especially python-lsp-server popped up multiple times with
vulnerabilities. Let's drop this.


rumpelsepp:
Really annoying and not affecting gallia.

image

PR #883, opened 2026-07-24
fix: Remove boltons dependency

rumpelsepp:
This one is only used at one single location and can be removed.


rumpelsepp:
@ferdinandjarisch There should be no change in behaviour :tm:. This dependency was used in building the CLI interface.

PR #885, opened 2026-07-27
Improve hr performance

rumpelsepp:
I tried coding with AI and I could optimize hr:

Replace three hot-path spots in PenlogRecord that run once per record
when hr reads a penlog file, with cheaper stdlib-only equivalents:

- PenlogPriority.to_level(): match-statement -> precomputed dict
  lookup. Also used by gallia's live console logging, not just hr.
- PenlogRecord.parse_json(): PenlogPriority(value) enum construction
  -> precomputed {value: member} dict, and six "key in dict" + subscript
  pairs -> .get(). Both are behavior-preserving (same ValueError on an
  invalid priority, same None-on-missing-field semantics).
- _format_record(): dt.strftime("%b %d %H:%M:%S.%f") -> manual f-string
  with a fixed month-abbreviation table. Behavior change: timestamps
  always render English month abbreviations now, instead of respecting
  the OS locale.

Benchmark: `hr -p trace` over a synthetic 2,000,000-record/420MB penlog
file, output redirected to /dev/null, best of repeated runs:

before: 11.10s
after:   8.85s  (~20% faster, byte-identical output aside from the
               intentional month-name change)

### DIFF
--- a/src/gallia/log.py
+++ b/src/gallia/log.py
@@ -183,23 +183,25 @@ class PenlogPriority(IntEnum):
 
     def to_level(self) -> Loglevel:
         """Converts an instance of PenlogPriority to :class:`Loglevel`."""
-        match self:
-            case self.TRACE:
-                return Loglevel.TRACE
-            case self.DEBUG:
-                return Loglevel.DEBUG
-            case self.INFO:
-                return Loglevel.INFO
-            case self.NOTICE:
-                return Loglevel.NOTICE
-            case self.WARNING:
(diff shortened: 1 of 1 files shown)


rumpelsepp:
@ferdinandjarisch any opinions?

rumpelsepp:
Superseded by #892, which reworks `PenlogReader` and `hr` and includes the `perf` commit of this PR unchanged. The `fix` commit is not needed anymore, because the code it fixes has been replaced there.

PR #887, opened 2026-09-02
fix(scan-identifiers): Catch errors when leaving session

ferdinandjarisch:
When leaving the session, the scan is basically done. Nonetheless, errors result in the entire scan to return with a non-zero return code. For single sessions this does not really make sense, and even for multiple sessions it should be enough to print a warning and let the remaining scan handle weird behavior of the ECU.


rumpelsepp:
Just wondering: Would that be something that might be useful in different places in the codebase? Or is this specific to this scanner?

ferdinandjarisch:
> Just wondering: Would that be something that might be useful in different places in the codebase? Or is this specific to this scanner?

Not sure what exactly you mean, but generally yes: there are parts of the code here and there which somewhat silently perform additional actions that deviate from the general behavior, e.g. by not handling all error cases in a similar matter.

To thoroughly check this, however, one might have to look at each individual scanner.

Concretely for `leave_session` I haven't checked whether other scanners have this exact same code pattern, honestly.

ferdinandjarisch:
I did a quick search and added a try block to the memory and reset scanner as well as to the seeds dumper.

PR #890, opened 2026-09-14
feat(discover-isotp): Add multiplier and offset for finer scans

ferdinandjarisch:
This commit adds `multiplier` and `offset` as CLI flags, which allow to easily modify the tested identifiers, e.g. to scan all identifiers `0x1ab`, `0x2ab`, `0x3ab`, `0x4ab`, `0x5ab`, ...


rumpelsepp:
Should we catch the case when the ID goes out of range for valid ISOTP IDs?

ferdinandjarisch:
> Should we catch the case when the ID goes out of range for valid ISOTP IDs?

If by "ISOTP IDs" you mean ISOTP's extended addressing, then this is already the case: https://github.com/Fraunhofer-AISEC/gallia/blob/fe62eae49fd66b3f7488cd168b3dd090214ab840/src/gallia/commands/discover/isotp.py#L110-L117

For CAN arbitration IDs I did some quick testing and feel this should be checked deeper in the stack anyhow, e.g. here: https://github.com/Fraunhofer-AISEC/gallia/blob/fe62eae49fd66b3f7488cd168b3dd090214ab840/src/gallia/transports/can.py#L64-L72

rumpelsepp:
No, I mean the `ID` variable in the proposed code:

``` python
        for _id, padding_byte in product(range(self.config.start, self.config.stop + 1), padding):
            ID = _id * self.config.multiplier + self.config.offset
            if self.config.extended_addressing is True:
                tx_id = self.config.tester_addr
                pdu = self.build_isotp_frame(req, ext_addr=ID, padding=padding_byte)  # <---- here
```

If in your code extended adressing is true, and multipier is set to, e.g. 100, then ID reaches 0xff after four loop iterations.

A quick check in ipython confirms that the `_build_isotp_frame_extended()` function crashes in this case. `ext_addr` goes through `build_isotp_frame()` unmodified.

```
_build_isotp_frame_extended([0xfe], 300)
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[5], line 1
----> 1 _build_isotp_frame_extended([0xfe], 300)

Cell In[3], line 2, in _build_isotp_frame_extended(pdu, ext_addr)
      1 def _build_isotp_frame_extended(pdu: bytes, ext_addr: int) -> bytes:
----> 2     isotp_hdr = bytes([ext_addr, len(pdu) & 0x0F])
      3     return isotp_hdr + pdu

ValueError: bytes must be in range(0, 256)
```

ferdinandjarisch:
Ah, now I understand!

Yes, I missed this, thanks for pointing it out; `multiplier` and `offset` were never supposed to apply to a scan of ISOTP's extended addressing, since search space is small enough for it.

I thus moved the assignment into the conditional, and additionally added a commit to properly raise an exception if any CAN arbitration ids surpass the 29 bit of extended addresses.

At a quick glance, ISOTP itself does not have this issue, simply because it (silently) masks away higher bits: https://github.com/Fraunhofer-AISEC/gallia/blob/9f16a044071cfacbd3aaf8aac0489c11dcbc6c4b/src/gallia/transports/isotp.py#L105-L109

PR #892, opened 2026-09-23
Merge cursed-hr into hr and rework PenlogReader

rumpelsepp:
This merges `cursed-hr` into `hr` (`hr --cursed`) and reworks the log reading on top of a new `PenlogReader`. It supersedes #885 and includes its performance commit.

## Changes

**`PenlogReader` (`gallia.log`)**
- Keeps only a compact index in memory (file offset and priority, nine bytes per record) and parses records on demand, so arbitrarily large logfiles can be read.
- The index is built incrementally and only as far as needed. Compressed files and pipes are decompressed chunk by chunk into a temporary file, so the first records are available immediately.
- `find()`/`rfind()` search by priority, and `search()`/`rsearch()` search the raw records with regexes without parsing them.
- Truncated files can be read up to the damaged part.
- New `stream_records()`/`stream_raw_records()` read a logfile sequentially with constant memory usage and no temporary file.
- **Breaking:** the stateful `seek_to_*()`, `current_record`, `current_priority`, and `readline()` APIs have been removed.

**`hr`**
- Streams logfiles, so output starts immediately and memory usage stays constant (~30 MB instead of the fully decompressed file in `/tmp`). Only `--reverse` needs random access.
- `-f/--filter` uses a simple query language instead of `eval()`, e.g. `module=scanner tag=result !timeout`. `hr --help` documents the syntax.
- Read errors, e.g. for truncated files, are reported after the readable records with exit code `IOERR`.
- New output options for both modes: `--no-prefix`, `--relative-timings` (relative to the first displayed record), and `--interpret` (UDS messages decoded as comments).
- Continuation lines of multiline records and stacktraces are aligned with the first line of the data, like in the viewer.
- Fixes: records with the priorities `EMERGENCY`/`ALERT` crashed `hr`; the colors depended on stderr instead of stdout being a terminal; and the next record continued the last line of a stacktrace.

**`hr --cursed`** (replaces `cursed-hr`, which had ~1450 lines with its own copy of the parsing logic)
- The first page is shown immediately (~0.12 s for a 1.2 GB logfile), and the rest is indexed in the background with asyncio. Progress is shown in the status line and in the terminal's native progress bar (OSC 9;4).
- Only the records around the visible part are kept in memory: ~60 MB for 3 million records.
- Priority zones and undo/redo work as before. Without a marked range, `p` now also works for showing *less*: it applies to the zone of the cursor (showing more still applies to the hidden records around the cursor). `p`/`P` report the changed records in the status line.
- New record view for debugging (`Enter`): decoded JSON with syntax highlighting, `data`/`stacktrace` with line breaks, the decoded UDS message, offset and size; `r` shows the exact raw record, and `n`/`N` step through the visible records.
- It looks exactly like plain `hr`: both share the formatting (`RecordFormatter`) and the colors of the log levels (`level_style()` in `gallia.log`), and neither duplicates the other.
- Mouse support (enabled by default, `--no-mouse` disables it): a click moves the cursor, a double click inspects the record, and the wheel scrolls.
- The record under the cursor is highlighted with a subtle background. Whether the terminal is dark or light is detected via OSC 11, with a DA1 query afterwards, so terminals without OSC 11 cause no delay; `--theme dark|light` overrides the detection.
- The status line shows the priority, timestamp, module, source line, and tags of the record under the cursor, plus the last key. While a key is being handled, a busy marker (`…`) shows that `hr` is working.
- The filter uses the new query language, and the help is an overlay.
- Adds support for terminal resizing, unicode input, stacktraces, and stdin.

**Logging**
- `logger.trace()`, `notice()`, and `result()` stored the line and function of the wrapper in `gallia/log.py` instead of the caller's; they now pass the `stacklevel` accordingly.

**From #885**
- `perf: Speed up penlog record parsing and formatting`, cherry-picked; the code is unchanged. The same fast timestamp formatting is now used in the viewer as well.
- The `fix` commit of #885 is not included, because the code it fixes has been replaced.

## Performance

3,000,000 records, 1.17 GB uncompressed (`.zst`), output to `/dev/null`:

| | `master` | this PR |
|---|---|---|
| `hr file \| head -1` | ~1.4 s | 0.10 s |
| `hr -p notice` | 7.5 s, ~0.9 GB | 6.7 s, ~30 MB |
| `hr -p trace` | 38.7 s, ~1.1 GB | 28.0 s, ~30 MB |
| `hr -t -n 10 -p trace` | 2.1 s | 2.3 s |
| `hr --cursed`, first page | several seconds (decompression + full index) | ~0.12 s |

The priority of the raw records is extracted without parsing them: `stream_raw_records()` has inline fast paths for `<N>{...` and `"priority": N`, and falls back to `_scan_priority()` (as used by the index of `PenlogReader`) for everything else, e.g. broken records, which are shown as `ERROR`. Both hot loops avoid enum accesses.

Effect of the #885 commit alone on top of this branch: 1.23–1.26× when many records are printed (`-p trace`, `-p notice`), and 1.03–1.06× when only a few are.

## Testing

- New pytest suites: `tests/pytest/test_log_reader.py`, `tests/pytest/test_hr.py`, and `tests/pytest/test_hr_tui.py`. The latter runs the viewer on a virtual screen and covers priorities, zones, marked ranges, undo/redo, the filter, the record view, colors, mouse, status line, cursor line, and theme detection.
- Mouse events (click, double click, wheel) were tested in tmux, which uses the X10 encoding.
- Additional bats tests for filtering, tail/reverse, and `--cursed`.
- Output is byte-identical to `master` for all modes (plain, `.zst`, `.gz`, stdin; `--head`, `--tail`, `--reverse`, priorities), except for the intended alignment of multiline records.
- The colors of the viewer are identical to plain `hr` (compared with the escape sequences of both on the same terminal).
- The viewer was tested interactively on large files, broken or truncated files, pipes, and empty files.
- Tested locally with Python 3.11–3.15, including 3.14t and 3.15t.


rumpelsepp:
Opening very large files now shows the content immediately. Loading of the rest happens in the background and is indicated in the status bar.

PR #893, opened 2026-09-23
Dissect logged messages according to their protocol

rumpelsepp:
**Stacked on #892**; only the last three commits are new (`feat(log): Log the protocol of the messages of transports` and the two following). It is a draft until #892 is merged.

This picks up the idea of #593: instead of guessing that hex data is UDS, the logfile says which protocol a message is, and `hr` dissects it with the dissector registered for this protocol.

## Changes

**Log format**
- New unstable field `_proto` with the protocol of the data of messages which transports read or write, e.g. `uds`. The protocols are named like the dissectors of Wireshark (`uds`, `iso15765`, `can`, `doip`, …).
- Only the user of a transport knows the protocol of the payload; it sets `BaseTransport.payload_proto`. The UDS client and server set `uds`, `RawCANTransport` logs CAN frames (`can`), and the ISO-TP discovery sets `iso15765` after sniffing the idle traffic (normal addressing only). The transport API is unchanged, so plugin transports keep working.
- The direction stays in the existing `read`/`write` tags.
- All transports log messages with the new `BaseTransport.log_io()` instead of copies of the same code. Thus, the ISO-TP transport tags read messages with `read` as well, the lines transports log the payload instead of the hex line with a trailing `0a`, and the HSFZ transport logs the payload like the DoIP transport.

**Dissectors (`gallia.dissect`)**
- A registry of dissectors by protocol; further dissectors can be registered with `@dissector("proto")`.
- UDS (moved from `hr`) and ISO-TP in CAN frames (single, first, consecutive frame, flow control; the payload of single frames is dissected as UDS).
- Records without protocol, e.g. of older logfiles, are dissected as UDS, if possible, like before.

**`hr`**
- `--interpret` is now `-d/--dissect`; the viewer toggles it with `d`. The filter supports the field `proto`, e.g. `proto=uds`.
- The record view (`Enter`) of the viewer shows the dissection and, if `tshark` is installed, the protocol tree of Wireshark.

## Wireshark

Wireshark can be used for single messages: the message is wrapped into a pcap file and piped into `tshark`, as exported PDU for the dissector named like the protocol, or as SocketCAN frame for CAN (ISO-TP via "Decode As", with UDS as payload). This takes ~0.1 s per call, which is fine for the record view, but too slow for every line of `hr`. Therefore, the one-line dissections are done by our own dissectors, and Wireshark is optional.

## Testing

- New tests for the dissectors, for the pcap format and `tshark` (skipped without `tshark`), for the log records of a transport (protocol, tags, and the caller's line), for `--dissect` and the `proto` filter, and for the dissection in the viewer.
- Each commit passes pytest, mypy, and ruff; the bats tests of `hr` pass.
- Tried with a generated logfile of a UDS scan and an ISO-TP discovery.


rumpelsepp:
I generated some fake data and opened it. Terminal resize works, the dissecting is a bit more capable than the version in master and you can open a log record via `enter` or double click to introspect a message. There wireshark is called to dissect (if available) it. The result is printed in the introspect box.

Bildschirmfoto vom 2026-09-24 07-09-02
Bildschirmfoto vom 2026-09-24 07-08-44
Bildschirmfoto vom 2026-09-24 07-09-16

